PR #16, opened 2016-09-12
use one of two og-image versions randomly

phillchill:
Update 2 new og-images for home page and let template choose one at random

WEBUMENIA-277


igor-kamil:
nice idea to use random og:image. unfortunately facebook cache them.
but at least for each route it will choose (and cache) a different one :)
would be nice to refresh it after deploy using http://developers.facebook.com/tools/debug
please use imageOptim for images and after that we can go :)

PR #18, opened 2016-10-06
media.webumenia.sk for serving image data for artworks

igor-kamil:
right now it will return image full url for requested id

### DIFF
--- a/app/Http/routes.php
+++ b/app/Http/routes.php
@@ -17,6 +17,18 @@ use App\Item;
 use App\Slide;
 use App\Order;
 
+Route::group(['domain' => 'media.webumenia.{tld}'], function () {
+    Route::get('/', function ($tld) {
+        return "webumenia media server";
+    });
+    Route::get('{id}', function ($tld, $id) {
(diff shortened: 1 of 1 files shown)


app/Http/routes.php
```diff
@@ -17,6 +17,18 @@
 use App\Slide;
 use App\Order;
 
+Route::group(['domain' => 'media.webumenia.{tld}'], function () {
```
igor-kamil: {tld} is there to support .local / .dev / ... for local development as well as .sk for production

phillchill: so what does {tld} do / stand for?

app/Http/routes.php
```diff
@@ -17,6 +17,18 @@
...
+    Route::get('/', function ($tld) {
+        return "webumenia media server";
+    });
+    Route::get('{id}', function ($tld, $id) {
+        $item = Item::find($id);
+        if ($item) {
```
phillchill: should we return a helpful error message if there is no item with that $id?

PR #22, opened 2017-04-24
Feature/moravska galerie import chunk

igor-kamil:


app/Http/Controllers/ImportController.php
```diff
@@ -189,94 +188,109 @@ public function launch($import, $file)
...
         $images = null;
         if ($import->dir_path) {
             $image_dir = basename($this_import_record->filename, '.csv');
             $images = \Storage::listContents('import/' . $import->dir_path . '/' . $image_dir);
         }
 
+
+        $file = (is_array($file)) ? storage_path('app/' . $file['path']) : $file;
+
+        $func = function ($row) {
```
phillchill: a more indicative function name would he helpful

app/Http/Controllers/ImportController.php
```diff
@@ -189,94 +188,109 @@ public function launch($import, $file)
...
+                $item->title = $row['Titul'];
+                $medium = ($row['MatSpec']) ? ($row['Material'] . ', ' . $row['MatSpec']) : $row['Material'];
+                $item->medium = $medium;
+                $technique = ($row['TechSpec']) ? ($row['Technika'] . ', ' . $row['TechSpec']) : $row['Technika'];
+                $item->technique = $technique;
+                $item->topic = ($row['Namet']) ? $row['Namet'] : '';
+                $item->inscription = $row['Sign'];
+                $work_type = Import::getWorkType($row['Rada_S'], $row['Skupina']);
+                $item->work_type = $work_type;
+                // este budu miry
```
phillchill: ??

app/Http/Controllers/ImportController.php
```diff
@@ -189,94 +188,109 @@ public function launch($import, $file)
...
-                                $iip_img = '/MG/DrevoHQ/jp2/'.$item_image_file['filename'].'.jp2'; // @TODO: change this!
-                                $iip_url = 'http://www.webumenia.sk/fcgi-bin/iipsrv.fcgi?DeepZoom='. $iip_img;
-                                if (isValidURL($iip_url)) {
-                                    $item->iipimg_url = $iip_img;
-                                }
+            foreach ($records as $index => $row) {
+
+                $row = array_map(function($value) {
+                   return $value === "" ? NULL : $value;
+                }, $row);
```
phillchill: I would separate the above anonymous functions into a named functions outside of the `foreach` (e.g. with that [other named function](https://github.com/SlovakNationalGallery/web-umenia-2/pull/22/files?diff=unified#diff-e489ef4664b1c24151b9b01896103403R200) ) and then simply call it here by name as parameter to `array_map()`

app/Http/Controllers/ImportController.php
```diff
@@ -189,94 +188,109 @@ public function launch($import, $file)
...
+                $identifier = $row['Rada_S'] . ' ' . (int)$row['PorC_S'];
+                $image_file = $row['Rada_S'] . str_pad($row['PorC_S'], 6, "0", STR_PAD_LEFT);
+
+                $suffix = ($row['Lomeni_S'] != '_') ? $row['Lomeni_S'] : '';
+                if ($suffix) {
+                    $id .= '-' . $suffix;
+                    $image_file .= '-' . $suffix;
+                    $identifier .= '/' .$suffix;
+                }
+
```
phillchill: I would continue the pattern started on the lines above of pulling data out of `$row` into variables (`$gallery`, `$id`, `$identifier`, etc) and separately construct an `$item` instance out of these variables, like you are doing on the [first lines below](https://github.com/SlovakNationalGallery/web-umenia-2/pull/22/files?diff=unified#diff-e489ef4664b1c24151b9b01896103403R239)
this way we help to separate two concerns:
* parsing the data from the csv, **involving logic**
* constructing the `$item` instance, **logicless**

app/Http/Controllers/ImportController.php
```diff
@@ -189,94 +188,109 @@ public function launch($import, $file)
...
+                    });
+                    if (!empty($item_image_files)) {
+                        $item_image_file = reset($item_image_files);
+                        $this->uploadImage($item, $item_image_file);
+                        $item->has_image = true;
+
+                        // detect DeepZoom at IIP - if it's hq (high quality) image
+                        if (strpos($item_image_file['filename'], 'hq') !== false) {
+                            $iip_img = '/MG/DrevoHQ/jp2/'.$item_image_file['filename'].'.jp2'; // @TODO: change this!
+                            $iip_url = 'http://www.webumenia.sk/fcgi-bin/iipsrv.fcgi?DeepZoom='. $iip_img;
```
phillchill: just checking, is this still correct?

app/Http/Controllers/ImportController.php
igor-kamil: just my note... that there would be also dimensions

app/Http/Controllers/ImportController.php
igor-kamil: you are right. I will fix this in the next task/branch - because I am going to improve parsing/mapping anyway

app/Http/Controllers/ImportController.php
igor-kamil: nice find!
right now yes. but of course this have be changed soon...

PR #33, opened 2017-11-08
Feature/map translated attrs in spice harvester controller

phivk:
I've published an overview of all attributes I worked on the wiki:
https://wiki.cedvu.sk/pages/viewpage.action?pageId=27983890


phivk:
> and include the changes into changelog

neither source branch `feature/map_translated_attrs_in_spiceHarvesterController` nor destination branch `feature/translatable-models from` currently have `CHANGELOG.md` (they are out of date with `develop`)

I'm afraid merging `develop` into both branches will cause merge conflicts in both branches, and then again when merging source to destination. I would prefer to do this merge cleanly and then merge `develop` into the updated `feature/translatable-models from` so that we are dealing with merge conflicts only 1x instead of 3x.

@igor-kamil what do you think?

PR #34, opened 2017-11-08
Feature/pattern lib

phivk:
first version of very basic pattern lib documentation of components


app/Http/routes.php
```diff
@@ -237,6 +237,44 @@ function()
...
     });
 
+    Route::get('patternlib', function () {
+        $COMPONENT_BASE_PATH = '../resources/views/components/';
+        
+        // get json file names
+        $file_names = array();
+        $handle=opendir($COMPONENT_BASE_PATH);
+        while (false !== ($file_name = readdir($handle))):
+            if(substr($file_name, -5) == '.json'):
```
igor-kamil: what is "-5" ?

app/Http/routes.php
```diff
@@ -237,6 +237,44 @@ function()
         return view('dielo', array('item' => $item, 'more_items' => $more_items, 'previous' => $previous, 'next' => $next));
     });
 
+    Route::get('patternlib', function () {
+        $COMPONENT_BASE_PATH = '../resources/views/components/';
```
igor-kamil: nice point to follow const convention with uppercase
why not to go even further and define it as const? -> `const COMPONENT_BASE_PATH` within class definitions

app/Http/routes.php
```diff
@@ -237,6 +237,44 @@ function()
         return view('dielo', array('item' => $item, 'more_items' => $more_items, 'previous' => $previous, 'next' => $next));
     });
 
+    Route::get('patternlib', function () {
```
igor-kamil: wouldn't it be nicer to use separated controller for `patternlib` ?

app/Http/routes.php
phivk: `substr($file_name, -5)` returns the last 5 character of the string $file_name
if this is equal to ".json" we know we are dealing with a JSON file

rastislav-chynoransky: there is more descriptive helper function in laravel named `ends_with`

app/Http/routes.php
```diff
@@ -237,6 +237,44 @@ function()
...
+        $COMPONENT_BASE_PATH = '../resources/views/components/';
+        
+        // get json file names
+        $file_names = array();
+        $handle=opendir($COMPONENT_BASE_PATH);
+        while (false !== ($file_name = readdir($handle))):
+            if(substr($file_name, -5) == '.json'):
+                $file_names[] = $file_name;
+            endif;
+        endwhile;
```
rastislav-chynoransky: this is just a detail (and we should definitely use linter in future), but please don't mix those two synatxes

app/Http/routes.php
```diff
@@ -237,6 +237,44 @@ function()
...
+        foreach ($file_paths as $file_path) {
+            $file_str = file_get_contents($file_path);
+            $file_json = json_decode($file_str, true);
+            $components_json[] = $file_json;
+        }
+
+        // add attributes
+        foreach ($components_json as $index => $component) {
+            $components_json[$index]['include_path'] = 'components.'.substr($file_names[$index], 0, -5);
+            $components_json[$index]['source_code'] = htmlspecialchars(file_get_contents(substr($file_paths[$index], 0, -5).'.blade.php'));
```
rastislav-chynoransky: is this `htmlspecialchars` necessary? isn't it what blade's {{ }} macro(?) does?

app/Http/routes.php
igor-kamil: try to keep following PSR-2 as coding standard - refer to http://www.php-fig.org/psr/psr-2/#11-example

igor-kamil:
please update the changelog following http://keepachangelog.com/en/1.0.0/

app/Http/routes.php
phivk: you're right:
https://laravel.com/docs/5.5/blade#displaying-data

PR #37, opened 2017-11-20
Feature/setup

igor-kamil:
fix seeders after upgrading L4.2 -> L5


database/seeds/DatabaseSeeder.php
```diff
@@ -11,14 +13,14 @@ public function run()
...
+		$this->call('HarvestsTableSeeder');
 		// $this->call('CollectionsTableSeeder');
-		// $this->call('CategoriesTableSeeder');
+		$this->call('CategoriesTableSeeder');
 		// $this->call('ArticlesTableSeeder');
-		// $this->call('RolesTableSeeder');
+		$this->call('RolesTableSeeder');
 		// $this->call('SketchbooksTableSeeder');
-		$this->call('SlidesTableSeeder');
+		// $this->call('SlidesTableSeeder');
```
phivk: why this precise selection of Seeders? 
I mean e.g. why is `HarvestsTableSeeder` enabled, but `CollectionsTableSeeder` not?

igor-kamil: the selection was made to distinguish which seeders you really need to setup webumenia on your own - and which are there just for some random data / or even leftovers f.e. `CollectionsTableSeeder` is coming from http://dvekrajiny.sng.sk/ - but it is still possible to use it as seeder )

PR #39, opened 2017-12-04
enable orders of reproductions with framing / mounting

igor-kamil:


ustengrav:
English translations for Christmas warnings

If you wish a reproduction with a passe-partout or a frame to be ready before Christmas, make sure to place your order before December 8th.

For a reproduction without a frame or a passe-partout to be ready before Christmas, place an order before December 15th.

PR #40, opened 2017-12-11
Tests setup

rastislav-chynoransky:


tests/DatabaseMigrations.php
```diff
@@ -0,0 +1,43 @@
...
+     *
+     * @return void
+     */
+    public function runDatabaseMigrations()
+    {
+        $this->parentRunDatabaseMigrations();
+
+        array_pop($this->beforeApplicationDestroyedCallbacks);
+
+        $this->beforeApplicationDestroyedCallbacks[] = function() {
```
rastislav-chynoransky: This is DB specific, but here we have to destroy DB, because migrations are currently broken.

phivk: > because migrations are currently broken.
wait... what does that mean?

rastislav-chynoransky: They won't rollback, because of an error

PR #41, opened 2017-12-11
Add importers and their tests

rastislav-chynoransky:


app/Traits/DatabaseMigrations.php
```diff
@@ -1,13 +1,13 @@
 <?php
 
-namespace Tests;
+namespace App\Traits;
```
rastislav-chynoransky: I couldn't load this class from tests directory. Not sure why

PR #45, opened 2017-12-20
Feature/color description

rastislav-chynoransky:


app/Console/Commands/DescribeItemColors.php
```diff
@@ -0,0 +1,66 @@
...
+
+use App\Descriptors\ColorDescriptor;
+use App\Item;
+use Illuminate\Console\Command;
+use Illuminate\Support\Facades\DB;
+use Illuminate\Support\Facades\Log;
+
+class DescribeItemColors extends Command
+{
+    protected $signature = 'color_descriptors:create {--all} {i?} {n?}';
```
phivk: not sure what i and n stand for. Might be nice to add [input descriptors](https://laravel.com/docs/5.3/artisan#input-descriptions)

app/Console/Commands/DescribeItemColors.php
```diff
@@ -0,0 +1,66 @@
...
+        $items = Item::query();
+
+        if (!$this->option('all')) {
+            $items->whereNull('color_descriptor');
+        }
+
+        $n = (int)$this->argument('n');
+        $i = (int)$this->argument('i');
+
+        if ($this->argument('n') === null && $this->argument('i') === null) {
```
phivk: better avoid the mixed usage of `$this->argument('n')` and `$n`. I'd stick with `$n` once you have assigned it on line 32.
Same for `$i`...

app/Console/Commands/DescribeItemColors.php
```diff
@@ -0,0 +1,66 @@
...
+            // go ahead
+        } else if ($this->argument('n') === null || $this->argument('i') === null) {
+            return $this->error("You must specify either none or both 'n' and 'i' arguments");
+        } else if ($i > $n || $i <= 0) {
+            return $this->error("Arguemnt 'i' must be less than 'n' and both have to be positive integers");
+        } else {
+            $items->where(DB::raw("created_at MOD $n"), '=', $i - 1);
```
phivk: what does this do?

app/Http/Controllers/ItemController.php
```diff
@@ -304,6 +304,10 @@ public function reindex()
         Item::chunk(200, function ($items) use (&$i) {
             $items->load('authorities');
             foreach ($items as $item) {
+                if (!$item->color_descriptor) {
+                    continue;
```
phivk: is this left over from debugging?

app/Http/routes.php
```diff
@@ -234,7 +234,52 @@ function()
...
+            }
+
+            $amount_sum = array_sum(array_column($colors_used, 'amount'));
+            foreach ($colors_used as $hex => $color_used) {
+                $colors_used[$hex]['amount'] = sprintf("%.3f%%", $colors_used[$hex]['amount'] * 100 / $amount_sum, 3);
+            }
+
+            uasort($colors_used, function ($a, $b) {
+                if ($a['amount'] == $b['amount']) {
+                    return-0;
```
phivk: should this be `return 0;`?

app/Http/routes.php
```diff
@@ -234,7 +234,52 @@ function()
...
         }
 
-        return view('dielo', array('item' => $item, 'more_items' => $more_items, 'previous' => $previous, 'next' => $next));
+        $similar_by_color = [];
+        $colors_used = [];
+
+        if ($item->color_descriptor) {
+            $similar_by_color = $item->similarByColor(100);
+
+            for ($i = 0; $i < count($item->color_descriptor) / 4; $i++) {
```
phivk: this logic for calculating `$colors_used` is quite a lot to write it all out in `routes.php`. Could it be refactored into a public method of `Item` so that this becomes:
`$colors_used = $item->colorsUsed($max_number_of_colors);`
?

app/Item.php
```diff
@@ -200,6 +204,45 @@ public function getOaiUrl()
...
+
+        if ($seconds < 0.001) {
+            $duration = round($seconds * 1000000) . 'μs';
+        } elseif ($seconds < 1) {
+            $duration = round($seconds * 1000, 2) . 'ms';
+        } else {
+            $duration = round($seconds, 2) . 's';
+        }
+
+        \Debugbar::info($duration);
```
phivk: leftover debugging?

app/Console/Commands/DescribeItemColors.php
rastislav-chynoransky: That was used just to distinguish between original input value (may be null) and converted to int:
`(int)null === 0`
But maybe it should be rewritten, if it is not understandable.

app/Console/Commands/DescribeItemColors.php
rastislav-chynoransky: This was easiest-to-implement "parallelization", to be able to run multiple instances of descriptor (because it takes a lot of time with our data)

rastislav-chynoransky: We might consider more elegant solution (pthreads?), but i've never done parallelization in PHP.

app/Http/Controllers/ItemController.php
rastislav-chynoransky: Yes. (Note: must check indexing null color_descriptor)

app/Http/routes.php
rastislav-chynoransky: 👍 And we could create Color model and pass that model around in our app

app/Http/routes.php
rastislav-chynoransky: I think, this one is ok

app/Console/Commands/DescribeItemColors.php
phivk: ah ok I see... In that case I have no clear rewrite in mind 👌

app/Http/routes.php
phivk: that would be even fancier 😉

app/Http/routes.php
phivk: I mean technically `return-0` and `return 0` are equivalent, but isn't that hyphen a typo?

rastislav-chynoransky: oh, i see. yeah, that's a typo

tests/ColorTest.php
```diff
@@ -0,0 +1,58 @@
...
+
+
+namespace Tests;
+
+
+use App\Color;
+
+class ColorTest extends TestCase
+{
+    public function testIncorrentInt() {
```
phivk: typo?
`testIncorrentInt()` instead of `testIncorrectInt()`

rastislav-chynoransky: You are really attentive!

phivk: I'm impressed myself that I spotted this 🔬

PR #55, opened 2018-02-23
restrict IIP depending on NG rights policy

igor-kamil:


app/Item.php
```diff
@@ -684,9 +676,15 @@ public function getRelationshipTypeAttribute($value)
     }
 
     public function getHasIipAttribute() {
+        if (!$this->is_iip_allowed) return false;
+
         return !$this->getZoomableImages()->isEmpty();
     }
 
+    public function getIsIipAllowedAttribute() {
```
rastislav-chynoransky: This could match the attribute name getHasRightsAttribute

igor-kamil: but rights are not talking only about IIP/zoom functionality. 
this is standalone check for zoom - regardless the `has_rights` attribute.

rastislav-chynoransky: Ok, you know better the business point of view.
But then what does has_rights mean?

PR #56, opened 2018-03-21
Feature/adjust reference strip in zoom

phivk:


app/Item.php
```diff
@@ -200,6 +200,16 @@ public function getOaiUrl()
         return Config::get('app.old_url').'/oai-pmh/?verb=GetRecord&metadataPrefix=oai_dc&identifier='.$this->id;
     }
 
+    public function getRelatedIIPImgUrls()
+    {
+        if (!empty($this->related_work)) {
+            return Item::where('related_work', '=', $this->related_work)->where('author', '=', $this->author)->whereNotNull('iipimg_url')->orderBy('related_work_order')->lists('iipimg_url')->toArray();
```
igor-kamil: you are already within `Item` object, so you can use `self::where(...` instead.

PR #62, opened 2018-04-23
WEBUMENIA-503 enable ArtworkCarousel in ArticlePage

phivk:
NB: TODO Philo to manually merge this branch into develop

### DIFF
--- a/resources/views/clanok.blade.php
+++ b/resources/views/clanok.blade.php
@@ -97,7 +97,18 @@ mapa??
 @stop
 
 @section('javascript')
-
+{!! Html::script('js/slick.js') !!}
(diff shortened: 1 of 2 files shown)


resources/views/clanok.blade.php
```diff
@@ -97,7 +97,18 @@
 @stop
 
 @section('javascript')
-
+{!! Html::script('js/slick.js') !!}
 <script type="text/javascript">
+    // carousel
+    $('.artworks-preview').slick({
```
igor-kamil: this should be refactored later on. to become a part of a component. 
however it doesn't block this PR to be merged. just keep it in mind @phivk .

PR #63, opened 2018-04-26
Add launching import via admin

igor-kamil:
just to note:
- new **migration**

to **listen the queue**, the suggested approach is 
```
php artisan queue:listen --tries=2
```

### DIFF
--- a/app/Jobs/ImportCsv.php
+++ b/app/Jobs/ImportCsv.php
@@ -50,10 +50,7 @@ class ImportCsv extends Job implements ShouldQueue
             return;
         }
 
-        $files = \Storage::listContents('import/' . $this->import->dir_path);
-        $csv_files = array_filter($files, function ($object) { return (isSet($object['extension']) && $object['extension'] === 'csv'); });
-
-        foreach ($csv_files as $file) {
+        foreach ($this->import->files as $file) {
             if (\App::runningInConsole()) {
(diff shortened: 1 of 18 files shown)


rastislav-chynoransky:
Please update `.env.example` to reflect expected `QUEUE_DRIVER` value

app/Jobs/ImportCsv.php
```diff
@@ -0,0 +1,80 @@
...
+    /**
+     * Execute the job.
+     *
+     * @return void
+     */
+    public function handle()
+    {
+        $import = $this->import; //temporary fix because of PHP5.6 and T_PAAMAYIM_NEKUDOTAYIM error
+
+        $import->status = $import::STATUS_IN_PROGRESS;
```
rastislav-chynoransky: Use classname to access its constant.

igor-kamil:
@rastislav-chynoransky thank you for code review. 
I fixed all the mentioned issues.
Except "view related methods to model". 
Well, I am also not fan of them.
Any better approach?

rastislav-chynoransky:
@igor-kamil The simplest solution i can think of is creating a function that is passed to the view together with other data. If the mapping is used anywhere else, only then I would create a global helper.

rastislav-chynoransky:
@igor-kamil Only one suggestion to views, do not use non-escaped output if not necessary

PR #74, opened 2018-06-01
refactored argument array in spiceharvester

Endut:
and modified create method of ItemImage to deal properly with uniqueness constraints


Endut:
rebased against master

PR #76, opened 2018-06-08
Proper multiple item images handling

rastislav-chynoransky:
Fixes WEBUMENIA-703

## Type of change

# Checklist:

- [x] I have updated the [CHANGELOG](https://github.com/SlovakNationalGallery/web-umenia-2/blob/dea47a352d25bbc406f7fcaed036d872cdced1da/CHANGELOG.md)
- [x] I have assigned someone to this PR


app/Http/Controllers/AuthorityController.php
```diff
@@ -86,18 +86,6 @@ public function store()
...
-    /**
-     * Display the specified resource.
-     *
-     * @param  int  $id
-     * @return Response
-     */
-    public function show($id)
-    {
-        $authority = Authority::find($id);
-        return view('authorities.show')->with('authority', $authority);
```
rastislav-chynoransky: This got no corresponding template

app/Http/Controllers/SpiceHarvesterController.php
```diff
@@ -901,25 +906,29 @@ private function parseDate($string)
 
     private function parseYear($string)
     {
-        return (int)end((explode('.', $string)));
+        $exploded = explode('.', $string);
+        return (int)end($exploded);
     }
 
-    private function resolveIIPUrl($iip_resolver)
+    private function resolveIIPUrls($iip_resolver)
```
rastislav-chynoransky: Is it ok to create more than one image per item? Are the data checked?

resources/views/authorities/show.blade.php
```diff
@@ -1,123 +0,0 @@
-<!DOCTYPE html>
```
rastislav-chynoransky: Duplicate of `resources/views/items/show.blade.php` (not used)

igor-kamil: good point. can be updated to display authority detail (would be nice). or just remove it.

app/Http/Controllers/SpiceHarvesterController.php
igor-kamil: yes. sure!

app/Http/Controllers/AuthorityController.php
igor-kamil: as I wrote below - you can either update `resources/views/authorities/show.blade.php` or remove that.
I would prefer the first option - can be done also just partially (not necessary to display everything and all relationships)

rastislav-chynoransky: I'd rather do it in different PR

app/Http/routes.php
```diff
@@ -184,15 +184,9 @@ function()
 
         $item = Item::find($id);
 
-        if (empty($item) || !$item->isFreeDownload()) {
+        if (empty($item) || !$this->publicDownload()) {
```
igor-kamil: should be `$item->publicDownload()` instead of `$this->publicDownload()`

app/Item.php
```diff
@@ -644,11 +635,6 @@ private function relatedAuthorityIds()
         return $ids;
     }
 
-    public function isFreeDownload()
```
igor-kamil: why to remove this method? you need to check this condition several times all around the code - now you have to use combination of `($item->isFree() && !$item->getZoomableImages()->isEmpty())` or `($item->isFree() && $item->has_iip)`
also there is a space to override the conditions for downloading the hi-res  image. what might be useful for other instances (e.g. to put in action custom "_rights_" etc...)

app/Http/routes.php
rastislav-chynoransky: Thanks!

PR #78, opened 2018-06-12
Feature/dockerize web umenia

Endut:
# Description

this PR adds dockerfiles and a docker-compose for the full stack of
- php/apache (web)
- mysql db (mysql)
- elasticsearch (elasticsearch)

to simplify management of dependencies and automatic building
start the dockerized web-umenia by first ```docker-compose build``` (this will take a while the first time)
then ```docker-compose up```

so the php part of webumenia lives in the container ```php```, therefore:  
run migrations with ```docker-compose exec php php artisan migrate --seed```
and setup elasticsearch with ```docker-compose exec php php artisan es:setup```
all other commands within the container of a service can be executed by prefacing with ```docker-compose exec <service name>```
in this initial dockerized setup port 9200 of elasticsearch container and port 3306 of the mysql container are exposed so you can still inspect what's going on in there from outside docker when doing local development (eg with kibana or some database app)

nb:
some files are touched that you might not expect -- eg files to do with connections to the elasticsearch service - this is so that the host and port at which web-umenia can speak to elasticsearch is read from the .env which can vary depending on whether you use docker or not. References to this value had to be tidied up a bit and made more consistent (so that 'https://' is not accidentally prepended to something that already has 'https://' in front)
this should still work consistently now without docker (elasticsearch is at 'https://localhost:9200'),
or with docker-compose (it will appear to web-umenia at 'http://elasticsearch:9200')

## Type of change

# Checklist:

- [x] I have updated the [CHANGELOG](/CHANGELOG.md)


.gitignore
```diff
@@ -30,3 +30,4 @@ caste_prikazy.txt
 public/sitemap*
 .env
 .env.testing
+data/
```
rastislav-chynoransky: Add leading slash, this will match `data` dir at any level. Yes, all directives above are wrong.

Dockerfile.elasticsearch2
```diff
@@ -0,0 +1,71 @@
+FROM ubuntu:14.04
```
rastislav-chynoransky: When to use this Dockerfile?

Dockerfile.elasticsearch
```diff
@@ -0,0 +1,62 @@
...
+RUN chmod +x gradlew ; \
+	./gradlew plugin ; \
+	mkdir -p $ES_HOME/plugins/elasticsearch-image-$ES_VERSION ; \
+	unzip build/distributions/elasticsearch-image-2.4.1.zip -d $ES_HOME/plugins/elasticsearch-image-$ES_VERSION	; \
+	rm -Rf $IMG_PLUGIN_SRC
+
+WORKDIR $ES_HOME
+# in our usual version of es (2.4.1) plugin install command is found as option to 'plugin' command in bin
+# later versions of elasticsearch have a seperate command plugin-install
+RUN ./bin/plugin install https://github.com/vhyza/elasticsearch-analysis-lemmagen/releases/download/v$ES_VERSION/elasticsearch-analysis-lemmagen-$ES_VERSION-plugin.zip ; \
```
rastislav-chynoransky: See https://github.com/SlovakNationalGallery/web-umenia-2/pull/60#discussion_r191394467

.gitignore
Endut: do you mean ./data/ and ./env?

Dockerfile.elasticsearch2
Endut: sorry we can get rid of it

Dockerfile.elasticsearch
Endut: do you mean about not needing to download *.lem files?
will double-check this later

.gitignore
rastislav-chynoransky: Without the dot

Endut: my feeling is that any data directory should not be added to version control, but yeah didn't know about this .gitignore peculiarity, will change it

rastislav-chynoransky: there could easily be data dir for tests in future

Dockerfile
```diff
@@ -0,0 +1,30 @@
+FROM php:5.6-apache
+
+COPY . /var/www
```
igor-kamil: please don't. it made me run out of free space. don't copy it. use a volume bound.
get inspired by https://www.elastic.co/guide/en/elasticsearch/reference/current/docker.html
```
Always use a volume bound on /usr/share/elasticsearch/data, as shown in the production example, for the following reasons:

The data of your elasticsearch node won’t be lost if the container is killed
Elasticsearch is I/O sensitive and the Docker storage driver is not ideal for fast I/O
It allows the use of advanced Docker volume plugins
```

Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+RUN apt-get update -y && apt-get install -y \
+	libmcrypt-dev \
+	libpng-dev
+
+RUN docker-php-ext-install \
+	mbstring \
+	pdo \
+	pdo_mysql \
+	gd \
+	mcrypt \
```
igor-kamil: add
```
    zip \
```

PR #101, opened 2018-09-07
Feature/webumenia 774 add vsg to the list of galleries

igor-kamil:
# Description

Add VSG to the list of galleries and allow download of public-domain artworks

Fixes [WEBUMENIA-774](https://jira.sng.sk/browse/WEBUMENIA-774)

## Type of change

# Checklist:


public/js/components/map-with-galleries.js
```diff
@@ -0,0 +1,26 @@
...
+  });
+
+
+  $("svg#map path.cls-4").hover(function() {
+    var classes = $( this ).attr('class').split(/\s+/);
+    $.each( classes, function( index, class_name ){
+        // if class_name is uppercased (e.g. SNG, OGD ...) it's gallery ID
+        if (class_name === class_name.toUpperCase()) {
+            console.log(class_name);
+            $(".galleries a#"+ class_name ).addClass('active');
```
phivk: since this is a regular HTML element, you can use `.toggleClass()`

phivk: also, Id prefer to use data attributes where we can, as they are more expressive and clear about what they are used for. So I'd propose to store `gallery_id` in `data-gallery` attributes for the `.galleries a` elements.

see my comment in `informacie.blade.php`

resources/views/informacie.blade.php
```diff
@@ -14,68 +14,28 @@
...
+            <div class="col-md-8 text-center">
+                {!! file_get_contents(public_path('images/mapa.svg')) !!}
             </div>
         </div>
+        <div class="row bottom-space galleries">
+            @foreach(array_chunk($galleries, ceil(count($galleries)/3)) as $chunk)
+                <div class="col-md-4">
+                    <ul class="list-unstyled lead">
+                    @foreach($chunk as $gallery)
+                        <li><a href="{!! URL::to($gallery['url']) !!}" id="{{ $gallery['id'] }}">{{ utrans($gallery['lang_string']) }}</a></li>
```
phivk: I'd prefer to store the gallery value in a data-attr:
```
<li><a href="{!! URL::to($gallery['url']) !!}" data-gallery="{{ $gallery['id'] }}">{{ utrans($gallery['lang_string']) }}</a></li>
```

public/css/less/custom.less
```diff
@@ -509,14 +509,31 @@ table a:not(.btn):not(.readmore-js-toggle):not(.no-border), .attributes a:not(.b
...
+        transition: fill 0.3s;
+        -webkit-transition-timing-function: cubic-bezier(0.2,1,0.3,1);
+        transition-timing-function: cubic-bezier(0.2,1,0.3,1);
+
+        &:hover, &.active {
+          fill: @primary-color;
+        }
+
+      }
+    }
```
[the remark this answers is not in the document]
phivk: to consider re UI: if we want the user to create a link between the hovered icon and the active underlined link, wouldn't it be better is icon and active link underline have the same color? 🤔

app/Http/routes.php
```diff
@@ -297,7 +297,68 @@ function()
     Route::get('informacie', function () {
         $items = Item::random(20, ['gallery' => 'Slovenská národná galéria, SNG']);
 
-        return view('informacie', ['items' => $items]);
```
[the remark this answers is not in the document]
igor-kamil: thanx :) 
in the future there should be standalone model representing galleries (probably also with db table) - but I guess that's outside of the scope of this task

public/js/components/map-with-galleries.js
```diff
@@ -0,0 +1,26 @@
+$("document").ready(function() {
+  $(".galleries a").hover(function() {
+    var gallery_id = $( this ).attr('id');
+    var old_classes = $("svg#map path."+gallery_id ).attr('class');
```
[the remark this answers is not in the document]
igor-kamil: thank you for pointing out the `classList()`. however working with classes on svg paths is little bit [tricky](https://stackoverflow.com/questions/36130050/add-class-on-svg-path-element) and I chose this - oldschool but backward compatible approach (working on IE as well)

resources/views/informacie.blade.php
igor-kamil: we use the gallery acronym as unique identifier everywhere around (CEDVU / webumenia / ...)
but in other cases data attribute is definitely right choice! 

right here we store gallery code for querying dom elements. that's what `classes` and `ids` are ideal for. data- attributes are better for storing additional (reusable) informations - but not for querying the element. to read more: https://intuio.at/blog/dont-use-data-attributes-to-find-html-elements-with-js/

public/css/less/style.less
```diff
@@ -21,4 +21,5 @@
 @import "color-list.less";
 @import "article-thumbnail.less";
 @import "artwork-carousel.less";
-@import "zoom-viewer.less";
\ No newline at end of file
+@import "zoom-viewer.less";
+@import "map-with-galleries.less";
```
phivk: @igor-kamil please stick to [component naming conventions](https://wiki.cedvu.sk/display/ODaZS/Component+Library+master+doc), so the name should be:

"TheGalleryMap"

Capitalisation and hyphenation of the actual file names (.less, .blade.php and .js) are currently not very consistent, but let's stick to what is already in place, so I propose:
`the-gallery-map.less`
`the-gallery-map.js`

PR #103, opened 2018-09-14
Feature/webumenia 785 reproductions page

phivk:
# Description

Fixes [#WEBUMENIA-785](https://jira.sng.sk/browse/WEBUMENIA-785)

## Type of change

- [X] New feature (non-breaking change which adds functionality)

# Checklist:

- [X] I have updated the [CHANGELOG](../CHANGELOG.md)
- [X] I have requested at least 1 reviewer for this PR


public/css/less/repro-offer.less
```diff
@@ -0,0 +1,5 @@
+.wu-repro-offer {
+    table td {
+        padding: 0 10px;
+    }
+}
```
igor-kamil: lets reuse what we have (till it is possible)
in [informacie.blade.php](https://github.com/SlovakNationalGallery/web-umenia-2/blob/master/resources/views/informacie.blade.php#L196-L198) we use already classes `table table-striped` for styling the <table> and getting the proper spacing. (in this case adding only class `table` looks even better imho)

I think the whole `repro-offer.less` can be removed for now

resources/lang/en/reprodukcie.php
```diff
@@ -0,0 +1,42 @@
...
+    'digital_examples'  => 'Príklady použitia digitálnych reprodukcii',
+    'digital_list'      => '<ul>
+                              <li>Každé poskytnutie digitálnej reprodukcie je <strong>podmienené podpísaním zmluvy</strong>, ktorá špecifikuje účel jej použitia.&nbsp;Odkaz na stiahnutie obrázkov zo servera SNG posielame až po obdržaní podpísanej zmluvy (odfotenej, oskenovanej alebo zaslanej poštou).</li>
+                              <li>Digitálne reprodukcie diel spravovaných inými galériami poskytujeme iba po predložení písomného súhlasu príslušnej inštitúcie. Názov galérie spravujúcej dotyčné dielo je uvedené v položke <strong>„Galéria“</strong> v informáciách o diele.</li>
+                              <li>Ak si chcete objednať aj tlačené aj digitálne reprodukcie, vytvorte <strong>zvlášť objednávku pre každý typ.</strong>       </li>
+                            </ul>',
+    'digital_choice'    => 'Vyberte si reprodukcie',
+    'more-items_button' => 'zobraziť všetkých'
```
igor-kamil: I guess this is going to be translated later ... when the SK text would be final... right?

resources/views/components/repro_offer.blade.php
```diff
@@ -0,0 +1,16 @@
+<div class="wu-repro-offer top-space bottom-space">
+    <h4 class="text-uppercase">{!! $title !!}</h4>
+    <img class="w-100 bottom-space" src="{{$img_url}}">
+    <div>{!! $description !!}</div>
+    <table>
```
igor-kamil: as noted above - I suggest adding the class `table` [bootstrap tables](https://getbootstrap.com/docs/3.3/css/#tables)

resources/views/reprodukcie.blade.php
```diff
@@ -0,0 +1,191 @@
...
+            <div class="col-xs-12">
+                @include('components.artwork_carousel', [
+                    'slick_target' => "artworks-preview",
+                    'items' => $items_digital,
+                ])
+            </div>
+        </div>
+        <div class="row content-section">
+            <div class="col-sm-12 text-center">
+                <a href="{!! URL::to('katalog?is_free=' . '1') !!}" class="btn btn-default btn-outline sans" >{{ trans('reprodukcie.more-items_button') }} <strong>{!! App\Item::forReproduction()->count() !!}</strong>  <i class="fa fa-chevron-right "></i></a>
```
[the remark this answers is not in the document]
rastislav-chynoransky: You may use https://github.com/SlovakNationalGallery/web-umenia-2/blob/1d9a289daef08e22a8b3f10f3d64930600d7630c/app/Item.php#L898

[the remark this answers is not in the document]
rastislav-chynoransky: What's a bit tricky is that the value is time-dependent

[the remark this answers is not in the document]
rastislav-chynoransky: We should consider saving the date from when the artwork is free instead, but that might be out of the scope of this issue.

resources/lang/en/reprodukcie.php
phivk: exactly. I will ask @ustengrav to update the language files.

resources/views/reprodukcie.blade.php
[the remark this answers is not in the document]
phivk: Thanks @rastislav-chynoransky 
this is how we are determining whether an item is free currently: 
https://github.com/SlovakNationalGallery/web-umenia-2/blob/1d9a289daef08e22a8b3f10f3d64930600d7630c/app/Item.php#L627-L646

Do you propose to improve on this 'heuristic' method? 
Definitely out of scope of this PR, but I'd like to hear your suggestions...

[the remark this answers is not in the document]
rastislav-chynoransky: Yes, sure, the current code is wrong.
As I said :top: just to save the expected date of free artwork and send the current date in query for comparison.

resources/views/reprodukcie.blade.php
```diff
@@ -0,0 +1,191 @@
...
+                        @include('components.repro_offer', [
+                            "title" => trans('reprodukcie.print_offer_standalone_title'),
+                            "img_url" => "/images/reprodukcie/format-1-samostatna.jpg",
+                            "description" => trans('reprodukcie.print_offer_standalone_description'),
+                            "pricing_options" => [
+                                [trans('reprodukcie.print_offer_until')." <strong>A4</strong>", "(21 x 29,7 cm)",   "30"],
+                                [trans('reprodukcie.print_offer_until')." <strong>A3</strong>", "(32,9 x 48,3 cm)", "40"],
+                                [trans('reprodukcie.print_offer_until')." <strong>A2</strong>", "(42 x 59,4 cm)",   "45"],
+                                [trans('reprodukcie.print_offer_until')." <strong>A1</strong>", "(59,4 x 84,1 cm)", "55"]
+                            ]
```
phivk: should this data live here in the reprodukcie template? 
or in `routes.php`? 
or elsewhere?

app/Http/routes.php
```diff
@@ -299,6 +299,12 @@ function()
 
         return view('informacie', ['items' => $items]);
     });
+
+    Route::get('reprodukcie', function () {
+        $items_print   = Item::random(20, ['gallery' => 'Slovenská národná galéria, SNG']);
+        $items_digital = Item::random(20, ['is_free' => true]);
```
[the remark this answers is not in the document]
igor-kamil: yes. it was just necessary to run `php artisan oai-pmh:download-images` before, to flag the attribute `has_image=true` for items

resources/views/components/nav_bar.blade.php
```diff
@@ -0,0 +1,56 @@
...
+            ])
+
+            <a class="navbar-brand no-border hidden-xs second-part" href="{!! URL::to('') !!}">
+                umenia
+            </a>
+        </div>
+
+        <div class="collapse navbar-collapse navbar-main-collapse">
+            <ul class="nav navbar-nav">
+                <li class="{!! (Request::is('katalog') || Request::is('dielo/*')) ? 'active' : '' !!}">
```
igor-kamil: I am happy to see this componetized as well 👍 !  please include it also in the changelog 😉 
I know that it's bit outside of this task - but we discussed it would be nice to include also "home" link into the navigation in collapsed version (displayed in mobiles only - when the logo is hidden)
Could you eventually include it?

public/css/less/custom.less
```diff
@@ -88,6 +88,13 @@ html {
 }
 
 
+/* Bootstrap 4 utilities */
+.px-0 {
+  padding-left: 0;
+  padding-right: 0;
+}
+
```
igor-kamil: what about placing these bootstrap4 utilities in standalone file - which could be easily removed after migrating to bootstrap4? (just an idea)

resources/views/reprodukcie.blade.php
igor-kamil: for now it's ok. but I guess later we would like to share these `pricing_options` also with `order form` and `section information` and then it would be nicer to have it stored in the model (it doesn't have to be in DB - can be just in mode)

resources/views/reprodukcie.blade.php
```diff
@@ -0,0 +1,191 @@
...
+<section class="">
+    <div class="container">
+        <div class="row top-space bottom-space">
+            <div class="col-xs-12 text-center grey">
+                <h3>{{ utrans('reprodukcie.digital_examples') }}</h3>
+            </div>
+        </div>
+        <div class="row">
+            <div class="col-xs-4">
+                <img class="w-100" src="/images/reprodukcie/digirepro-1.jpg">
```
igor-kamil: it's better to use class [img-responsive](https://getbootstrap.com/docs/3.3/css/#images-responsive) in this case instead of `w-100`. also you can make extra step and include also class `lazyload` and rename `src=""` into `data-src=""` to use lazyloading. same as here https://github.com/SlovakNationalGallery/web-umenia-2/blob/develop/resources/views/components/article_thumbnail.blade.php#L4
please do the same for `` in `repro_offer.blade.php`

resources/views/components/nav_bar.blade.php
phivk: there is a separate issue for that already
https://jira.sng.sk/browse/WEBUMENIA-711

resources/views/reprodukcie.blade.php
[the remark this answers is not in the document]
phivk: @rastislav-chynoransky thanks for the suggestion, I agree. I have created this issue for it:
https://jira.sng.sk/browse/WEBUMENIA-833

phivk:
@igor-kamil I think I have addressed all your points. 
could you review again?

PR #109, opened 2018-10-19
Fix/1.6.0 spice harvester fix

igor-kamil:
# Description

Fix spice harvester for Authorities.
Since roles field is casted as array it should be also stored as array.
Also events start_date and end_date were not harvested properly. 
And also not stored properly - since the database used `DATE` type which doesn't support years from before `1000-01-01`.

## Type of change

# Checklist:


app/Harvest/Mappers/AbstractMapper.php
```diff
@@ -48,7 +48,8 @@ public function map(array $row) {
      * @param mixed $value
      */
     protected function setMapped(array &$mapped, $key, $value) {
-        if (is_array($value)) {
+        // dont serialize roles - they are already casted as array
+        if (is_array($value) && !str_contains($key, 'roles:')) {
```
igor-kamil: @rastislav-chynoransky this could be probably written in nicer way. any suggestions?

igor-kamil:
@rastislav-chynoransky however the biggest problem is still, that all `hasMany` relationships are causing duplications. method `processHasMany()` behaves bit weird.
also it would be more then nice to have option for some attributes to always replace hasMany relationships (especially for authorities) - to cover also when some event/name/.. is deleted in CEDVU.

app/Harvest/Mappers/AbstractMapper.php
rastislav-chynoransky: (Accidentally resolved) I'd rather get rid of automatic serialization. It's left over from old harvester.

rastislav-chynoransky:
> @rastislav-chynoransky however the biggest problem is still, that all `hasMany` relationships are causing duplications. method `processHasMany()` behaves bit weird.
> also it would be more then nice to have option for some attributes to always replace hasMany relationships (especially for authorities) - to cover also when some event/name/.. is deleted in CEDVU.

What entities are being duplicated? For each of them there are defined attributes to match against and should be easy to modify.
Items are being deleted by specifing such request via status attribute. Do we want to delete relationships when not present in export?

igor-kamil:
> What entities are being duplicated? 

e.g. `names` and `events` in `Authority`

> For each of them there are defined attributes to match against and should be easy to modify.
Items are being deleted by specifing such request via status attribute. Do we want to delete relationships when not present in export?

eventually yes. e.g. for `names` & `events` definitely. the only source is from harvest / import. it's not possible to set them manually. they should stay 100% identical as in the source.

igor-kamil:
@rastislav-chynoransky good point. thank you for being on alert :)
well.. your code seems to be right. but it behaves weird - often in returns null even if condition "should pass". and creates another record. (but not always).
I tried this - and it works exactly as expected.
Beside the fact that it will ignore non-fillable attributes (what might be problem). But if you keep all "fillable" to be fillable - then it works fine.

Anyway - @mcudrnak said, that the best scenario would be to replace the data anyway (to be able to also remove from relationship). so I will probably update it in that way. (or option to specify `relationships_to_always_replace`)

Your suggestions are more then welcome.

Issue #36, opened 2017-11-20
refactor list of json files within directory using Laravel Filesytem


igor-kamil:
https://github.com/SlovakNationalGallery/web-umenia-2/blob/bd094266652a33f21d1b9715cdea93c848c8fd9d/app/Http/routes.php#L244-L251

refactor using Laravel Filesytem or any nicer way

PR #132, opened 2018-12-14
Clear out patternlib controller

Kyslik:
# Description

Take on issue #36, I started from this [exact version](https://github.com/SlovakNationalGallery/web-umenia-2/blob/bd094266652a33f21d1b9715cdea93c848c8fd9d/app/Http/routes.php#L244-L251) and did not notice https://github.com/SlovakNationalGallery/web-umenia-2/commit/a0fc65164883e03ce1f8e556f37b78709c3a65ab. Anyway I think I improved already improved version.

- takes care of #36
- uses [flysystem](https://flysystem.thephpleague.com/) as requested
- uses fewer foreach loops (worst case: 2)

## Type of change

- [x] none below; refactoring I'd say

# How I tested the change

Since there are no tests for this functionality I did not bother installing the app as is and used L5.7 dummy project instead. I did return [key sorted array](https://stackoverflow.com/a/4501406/1564365) without a view:

```
recur_ksort($components); // https://stackoverflow.com/a/4501406/1564365
return response($components);
```

And compared `shasum` of original code and improved code:

```
➜  dummy curl -s http://dummy.test/original-master | shasum
5def6ea02f78d31dcc44a8ce8220ede8393afb1f  -
➜  dummy curl -s http://dummy.test/improved | shasum
5def6ea02f78d31dcc44a8ce8220ede8393afb1f  -
➜  dummy
```

### DIFF
--- a/app/Http/Controllers/PatternlibController.php
+++ b/app/Http/Controllers/PatternlibController.php
@@ -2,77 +2,69 @@
 
 namespace App\Http\Controllers;
 
-class PatternlibController extends Controller
+use League\Flysystem\Adapter\Local;
+use League\Flysystem\Filesystem;
+
+class ImprovedController extends Controller
 {
-    const COMPONENT_BASE_PATH = '../resources/views/components/';
+
+    /** @var \League\Flysystem\Filesystem */
+    protected $fs;
+
+
+    public function __construct()
+    {
+        $adapter  = new Local(resource_path('views/components'));
+        $this->fs = new Filesystem($adapter);
+    }
+
 
     public function getIndex()
-    {            
-        // get json file names
-        $file_names = array();
-        $handle = opendir(self::COMPONENT_BASE_PATH);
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/PatternlibController.php
```diff
@@ -2,77 +2,69 @@
...
             }
         }
-        return $component;
-    }
 
-    private function hydrateIncludePathJs($component)
-    {
-        if ( isset($component['include_js']) ) {
-            $component['include_path_js'] = $component['include_path'].'_js';
+        if (isset($array['include_js']) && (bool)$array['include_js'] === true) {
```
Kyslik: `bool` casting may give out false hopes; I naïvely thought that it would work in case of incorrect json (eg: `"include_js": "false"`)

```
>>> var_dump((bool)"true")
bool(true)
=> null
>>> var_dump((bool)"false")
bool(true)
=> null
```

PR #139, opened 2018-12-17
Feature/khb gallery in authority detail

igor-kamil:
# Description

Display gallery in author detail.
Using plugins:

- https://isotope.metafizzy.co/ (for masonry layout)
- http://dimsemenov.com/plugins/magnific-popup/ (for popup functionality with gallery)

Fixes WEBUMENIA-872

## Type of change

# Checklist:


resources/views/autor.blade.php
```diff
@@ -131,59 +160,53 @@
...
 @stop
 
-
 @section('javascript')
-{!! Html::script('js/readmore.min.js') !!}
-
-{!! Html::script('js/slick.js') !!}
-{!! Html::script('js/components/artwork_carousel.js') !!}
+<script src="https://unpkg.com/isotope-layout@3/dist/isotope.pkgd.min.js"></script>
+<script src="https://cdnjs.cloudflare.com/ajax/libs/magnific-popup.js/1.1.0/jquery.magnific-popup.min.js"></script>
```
phivk: would it be preferable to have these files locally?

resources/views/autor.blade.php
```diff
@@ -77,12 +77,41 @@
...
                         'show' => true,
                     ])
-                    @include('components.khb_accordion_card', [
-                        'title' => utrans('autor.gallery'),
-                        'content' => '',
-                        'parrentId' => 'authorAccordion',
-                        'show' => true,
-                    ])
+
+                    {{-- gallery --}}
```
igor-kamil: @phivk sorry I know this is not nice. would be better to do it somehow with combination of @yield directive... but that's probably for later

resources/views/layouts/master.blade.php
```diff
@@ -54,6 +54,8 @@
     </div>
 
     <!-- Javascript -->
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/lazysizes/4.1.1/lazysizes.min.js"></script>
```
phivk: same question here re local vs CDN...?

resources/views/autor.blade.php
igor-kamil: if we can use CDN - it's better to use CDN

https://www.sitepoint.com/7-reasons-to-use-a-cdn/
https://webmasters.stackexchange.com/questions/92083/local-files-vs-cdn

it really depends. usually each plugin has its own requirements and it's nice to let it be solved by npm. but it's not necessary in this case. and we don't customize these files.

PR #145, opened 2018-12-18
Feature/khb random leftover fixes

igor-kamil:
# Description

see commits

## Type of change

# Checklist:


resources/views/autori.blade.php
```diff
@@ -76,24 +76,24 @@
 </section>
 
 <section class="purpose py-5">
-    <p>Zámer</p>
-    <p>V roku 2019 plánujeme postupne pridávať do databázy ďalších umelcov.<br>
-    Zapíšte sa do nášho newslettra a dostávajte pravidelné info!</p>
+    <p>{{ trans('autori.intention') }}</p>
+    <p>
+        @if (LaravelLocalization::getCurrentLocale() == 'sk')
```
phivk: why is this in a conditional statement?
why is it not in a language file?

resources/lang/en/autori.php
```diff
@@ -15,4 +15,6 @@
     'authors_found'       => 'artists found for',
     'authors_counted'     => 'artists',
     'authors_none'        => 'currently no artists',
+    'intention'                 => 'Intention',
+    'subscribe'           => 'If you want to receive updates about Artbase, please subscribe to our newsletter.',
```
phivk: another day I'd have feedback on this, it's OK for now 👍

resources/views/autori.blade.php
igor-kamil: because I am stupid :D 
because they don't have `en` translation and for `en` version they don't want to even display it

but you are right, that way cleaner solution is to define empty string in `en` language file

also . because I didn't want to search how to define translation with parameter
but this is something I can fix/improve

PR #151, opened 2019-01-16
Hotfix/1.6.2

igor-kamil:
# Description

Fix search autocomplete for Collections and Articles after making them translatable.
Fix division by zero exception when counting aspect ratio for corrupted images.
Fix work_type filter in catalogue view.

Fixes WEBUMENIA-888, WEBUMENIA-864, WEBUMENIA-894

## Type of change

# Checklist:

### DIFF
--- a/resources/views/components/artwork_carousel.blade.php
+++ b/resources/views/components/artwork_carousel.blade.php
@@ -20,7 +20,7 @@
 
     @php
         list($width, $height) = getimagesize(public_path() . $item->getImagePath());
-        $width = $width * ($size / $height);
+        $width = $width * ($size / max($height,1));
         $height = $size;
     @endphp
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,8 @@ All notable changes to this project will be documented in this file[^1].
 
 ## [1.6.2] - 2019-01-16
 ### Fixed
-- autocomplete for Collections and Articles
+- search autocomplete for Collections and Articles
+- division by zero exception for corrupted images
 
 ## [1.6.1] - 2018-12-06
 ### Added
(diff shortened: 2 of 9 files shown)


resources/views/components/artwork_carousel.blade.php
```diff
@@ -20,7 +20,7 @@
 
     @php
         list($width, $height) = getimagesize(public_path() . $item->getImagePath());
-        $width = $width * ($size / $height);
+        $width = $width * ($size / max($height,1));
```
phivk: I'd prefer to just apply this `max()` trick once, within this `@php` block, so that the rest of the template can stay 'dumb' and simply use `$height` and `$width`.

I'd go for something like:
```
@php
    list($imgWidth, $imgHeight) = getimagesize(public_path() . $item->getImagePath());
    $width = max($imgWidth, 1) * ($size / max($imgHeight, 1));
    $height = $size;
@endphp
```
and then the rest of the template can stay as it is, without `max($width, 1)`

phivk: sorry, I must have been confused and thought you were using `max()` in multiple places in `artwork_carousel.blade.php`... I see now that you are not 🙂 

I did find out another thing in `dielo.blade.php`: it seems that not all instances of `$width` are encapsulated by `max()` so it looks like this could still result in 500 errors.

https://github.com/SlovakNationalGallery/web-umenia-2/blob/0ba2e0a48f004c99ea299ae481d05bdeddd26d07/resources/views/dielo.blade.php#L53-L63

I would propose to do all `max()` calculation inside the `php` blocks in all templates (like `dielo.blade.php`, etc) and then simply use `$width` in styling blocks etc as it currently is.

PR #153, opened 2019-01-23
Feature/webumenia 809 disable prints for non zoom

igor-kamil:
# Description

The artworks which don't have zoom images won't have the option to order printed reproduction, only digital.

Fixes WEBUMENIA-809

## Type of change

# Checklist:

### DIFF
--- a/app/Http/routes.php
+++ b/app/Http/routes.php
@@ -66,7 +66,18 @@ function()
 
         $items = Item::with('images')->find(Session::get('cart', array()));
 
-        return view('objednavka', array('items' => $items));
+        $allow_printed_reproductions = true;
+
+        foreach ($items as $item) {
+            if (!$item->hasZoomableImages()) {
+                $allow_printed_reproductions = false;
+            }
+        }
+
+        return view('objednavka', [
+            'items' => $items,
+            'allow_printed_reproductions' => $allow_printed_reproductions,
+        ]);
     });
 
     Route::post('objednavka', function () {
--- a/resources/lang/en/objednavka.php
+++ b/resources/lang/en/objednavka.php
@@ -37,6 +37,7 @@ return array(
     'form_format_for-download'     => 'For download:',
     'form_format_digital'          => 'digital reproduction',
     'form_purpose-alert'           => 'The copyright law prevents us from providing digital reproductions of <abbr title="less than 70 years has followed the death of the artist" data-toggle="tooltip">copyright protected artworks</abbr> for general private purposes (decoration, gift etc.). Slovak National Gallery will create a written contract, according to the information provided in the order. This contract will limit the use of the reproduction just for the purpose stated in the order.',
+    'form_purpose-alert-print'     => '',
     'form_purpose-label'           => 'Purpose',
     'form_purpose-info'            => 'Purpose description',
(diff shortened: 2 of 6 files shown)


resources/lang/en/objednavka.php
```diff
@@ -37,6 +37,7 @@
     'form_format_for-download'     => 'For download:',
     'form_format_digital'          => 'digital reproduction',
     'form_purpose-alert'           => 'The copyright law prevents us from providing digital reproductions of <abbr title="less than 70 years has followed the death of the artist" data-toggle="tooltip">copyright protected artworks</abbr> for general private purposes (decoration, gift etc.). Slovak National Gallery will create a written contract, according to the information provided in the order. This contract will limit the use of the reproduction just for the purpose stated in the order.',
+    'form_purpose-alert-print'     => '',
```
phivk: why is this EN string empty?

app/Http/routes.php
```diff
@@ -66,7 +66,18 @@ function()
...
+        foreach ($items as $item) {
+            if (!$item->hasZoomableImages()) {
+                $allow_printed_reproductions = false;
+            }
+        }
+
+        return view('objednavka', [
+            'items' => $items,
+            'allow_printed_reproductions' => $allow_printed_reproductions,
+        ]);
```
phivk: `$allow_printed_reproductions`  is set to `false` if one item in `$items` doesn't have zoomable images. This prevents other items in `$items` (who do have zoomable images) from being ordered as printed reproductions.

This seems like incorrect behaviour...

resources/views/objednavka.blade.php
```diff
@@ -126,6 +138,9 @@
 <div id="ucel">
     <div class="alert alert-info col-lg-offset-2 col-md-offset-4" role="alert">
         {!! trans('objednavka.form_purpose-alert') !!}
+        @if ($allow_printed_reproductions)
+            <br><strong>{!! trans('objednavka.form_purpose-alert-print') !!}</strong>
+        @endif
```
phivk: why display this `form_purpose-alert-print` sentence in case `$allow_printed_reproductions` is `true`?

resources/lang/en/objednavka.php
igor-kamil: I had to split `form_purpose-alert` into two parts - because in Slovak translation the second part didn't make sense when only digital reproduction was allowed.
But this sentence never existed in EN translation. So there is just empty string for such a key. 
But maybe it makes sense to add it cc @ustengrav ?

app/Http/routes.php
igor-kamil: It is proposed behaviour (not saying I like it)

resources/views/objednavka.blade.php
igor-kamil: it is displayed only when you choose digital reproduction. and it contains:
```
'V prípade záujmu o tlač výtvarných diel môžete využiť objednávku na tlačenú reprodukciu, kde výrobu a úpravu výtlačku zabezpečuje SNG.',
```
When person choose digital reproduction - we remind him "consider also printed reproduction".
But it makes no sense if we don't allow printed options (when you have non-digitised artwork in order)

PR #155, opened 2019-01-23
Fix/khb-webumenia 880 en missing artist birthplace

Endut:
# Description
adds separate form fields in admin for birth place (sk) and birth place (en), this way as long as the khb curators update the birth place for each locale using the admin (eg 'Londyn' and 'London') the birth place can be displayed in authority detail

Fixes #880 (link to Jira or GitHub issue)

## Type of change

# Checklist:

### DIFF
--- a/resources/views/authorities/form.blade.php
+++ b/resources/views/authorities/form.blade.php
@@ -70,7 +70,7 @@
             <div role="tabpanel" class="tab-pane  {{ ($i==0) ? 'active' : '' }}" id="{{ $locale }}">
               <div class="form-group">
                 {!! Form::label($locale . "[birth_place]", 'Miesto narodenia '.strtoupper($locale)) !!}
-                {!! Form::text($locale . "[birth_place]", Input::old($locale . "[birth_place]"), array('class' => 'form-control')) !!}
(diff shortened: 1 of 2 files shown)


resources/views/authorities/form.blade.php
```diff
@@ -68,7 +68,10 @@
           <div class="tab-content top-space">
             @foreach (\Config::get('translatable.locales') as $i=>$locale)
             <div role="tabpanel" class="tab-pane  {{ ($i==0) ? 'active' : '' }}" id="{{ $locale }}">
-
+              <div class="form-group">
+                {!! Form::label($locale . "[birth_place]", 'Miesto narodenia '.strtoupper($locale)) !!}
+                {!! Form::text($locale . "[birth_place]", Input::old($locale . "[birth_place]"), array('class' => 'form-control')) !!}
```
igor-kamil: there should be `@$authority->translate($locale)->birth_place : ''` 
instead of `Input::old($locale . "[birth_place]")`
in this way it doesn't display anything

Endut: aha, `Input::old` is only used in the case that a form is returned with errors?

PR #171, opened 2019-02-13
document style compilation

phivk:
# Description

Document the process of compiling LESS to CSS properly in Readme. I also removed rogue style(s).css files.

Fixes # https://jira.sng.sk/browse/WEBUMENIA-917

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

# Checklist:

- [X] I have updated the [CHANGELOG](../CHANGELOG.md)
- [X] I have requested at least 1 reviewer for this PR
- [X] I have made corresponding changes to the documentation


public/css/less/style.less
```diff
@@ -1,5 +1,3 @@
-// out: ../styles.css, compress: true
```
igor-kamil: this is/was here for the sublime plugin. 
also it's useful to know, where to store store the compiled CSS.
better to just fix the path (`style.css` instead of `styles.css`) and keep it.

PR #181, opened 2019-02-26
WEBUMENIA-862 Restrict bots from color searching

rastislav-chynoransky:
Fixes https://jira.sng.sk/browse/WEBUMENIA-862

## Type of change

# Checklist:

### DIFF
--- a/resources/views/dielo.blade.php
+++ b/resources/views/dielo.blade.php
@@ -306,20 +306,14 @@
     </div>
 </section>
 
-@if ($colors_used || $similar_by_color)
+@if ($colors_used)
 <section class="content-section">
     <div class="container">
         <div class="row">
-            <div class="col-xs-12">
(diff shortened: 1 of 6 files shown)


resources/views/dielo.blade.php
```diff
@@ -418,6 +412,14 @@ function rightArrowPressed() {
...
     $(document).ready(function(){
 
+        var colorRelated = $('#colorrelated');
+        if (colorRelated) {
+            var fetchUrl = colorRelated.data('fetch-url');
+            $.get(fetchUrl, function (data) {
+                colorRelated.append(data);
+            });
+        }
+
```
igor-kamil: very nice idea to load color related asynchronously. however - it would be nice to add some kind of loader + keep that space already "expanded" - to avoid jumping of the content while loading. 
anyway - that's another issue. for now this is just fine.

rastislav-chynoransky: Yeah, we can consider this feature as experimental and see how it affects the performance

PR #192, opened 2019-03-22
WEBUMENIA-939 Fix failing harvesters on zero results

rastislav-chynoransky:
Fixes https://jira.sng.sk/browse/WEBUMENIA-939

## Type of change

# Checklist:


app/Harvest/Repositories/AbstractRepository.php
```diff
@@ -43,11 +44,15 @@ public function getRows(SpiceHarvesterHarvest $harvest, \DateTime $from = null,
...
-        $total = $records->getTotalRecordCount();
+        try {
+            $total = $records->getTotalRecordCount();
 
-        foreach ($records as $record) {
-            $row = $this->getDataRecursively($record, $this->fieldMap);
-            yield $row[0];
+            foreach ($records as $record) {
+                $row = $this->getDataRecursively($record, $this->fieldMap);
+                yield $row[0];
```
igor-kamil: what will be returned by this function if `$records` is empty?

rastislav-chynoransky: it always returns a `Generator` https://www.php.net/manual/en/language.generators.overview.php

rastislav-chynoransky: > even if there are no more records to harvest, it should change status from `queued` to `completed`
> (now it stays in `queued` if there are no records found)
> 
> it's easy to emulate that scenario when you set `start_date` too early... e.g.
> 
> ```
> php artisan oai-pmh:harvest --start_date=2019-06-01
> ```

@igor-kamil `Phpoaipmh\RecordIterator::getTotalRecordCount` always throws `Phpoaipmh\Exception\MalformedResponseException` for me when there are zero results. Are you sure it doesn't change the status? It works for me. I can imagine that it won't switch it when it dies on fatal error, but I wasn't able to reproduce it.

PR #202, opened 2019-06-10
[docker] Test env

rastislav-chynoransky:
I had to update compose file format version to 2.2 because we want to set env vars for tests, but they are already set for container and cannot be easily changed. This may be fixed in future by upgrading to PHPUnit>=6.3 (requires PHP>=7) where we can force env vars in `phpunit.xml` config https://github.com/sebastianbergmann/phpunit/pull/2723 or by creating whole new test environment with separated containers.

# Type of change

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file[^1].
 ## [Unreleased]
 ### Added
 - map all datings in item harvester
+- docker test runner
 
 ### Added
 - Improved reproduction images and enabled ReproductionsPage in NavBar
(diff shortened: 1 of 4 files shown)


rastislav-chynoransky:
@igor-kamil please, test w/ mac. thanks!

igor-kamil:
> @igor-kamil please, test w/ mac. thanks!

@rastislav-chynoransky  it doesn't work for me.
returns `ERROR: No such service: #` for `docker-compose exec $(sed -e 's/^/-e /' .env.testing) php ./vendor/bin/phpunit "$@"`

however I still used docker 2.0. do you use the newer version?

PR #236, opened 2019-10-01
Feature/webumenia 666 frontend fix2

igor-kamil:
# Description

Fix handling of multiple images in artwork detail.
Bring back the lazyloading + responsive images (using srcset).
Add lazyloading functionality to `patternlib`

Fixes [WEBUMENIA-666](https://jira.sng.sk/browse/WEBUMENIA-666) [WEBUMENIA-770](https://jira.sng.sk/browse/WEBUMENIA-770)

## Type of change

# Checklist:


resources/views/components/image_carousel.blade.php
```diff
@@ -1,13 +1,25 @@
...
     <div class="artwork-carousel {{$slick_target}} {{$slick_variant or ''}}">
-        @foreach ($img_urls as $img_url)
-            <a href="{{$anchor_href}}" data-toggle="tooltip" data-placement="top" title="{{$anchor_title}}">
-                <img
-                    src="{{$img_url}}"
-                    class="mw-100 lazyload position-relative"
-                    alt="{!! $img_title !!} "
-                >
-            </a>
+        @foreach ($images as $image)
```
eronisko: Is the indentation here OK?

resources/views/components/image_carousel.blade.php
```diff
@@ -1,13 +1,25 @@
...
+
+        @php
+            list($width, $height) = getimagesize($image->getPreviewUrl());
```
igor-kamil: this is actually problem https://www.php.net/manual/en/function.getimagesize.php#88793
> getimagesize will download the entire image before it checks for the requested information. This is extremely slow on large images that are accessed remotely. Since the width/height is in the first few bytes of the file, there is no need to download the entire file.

I am now deciding between storing width/height in cache (so it will be loaded just once - can be nicely handled within `\App\ItemImage`) or storing it directly in the DB (thus having it right after retrieving the model instance).

Any suggestions? @rastislav-chynoransky @eronisko @martin-havala

eronisko: My 2c would be to store the dimensions in database.
Another option might be to estimate (or choose) the size and let JavaScript/browser handle any discrepancies.

eronisko:
@igor-kamil is this ready for review?

eronisko:
Closing this since it's been sitting here for over a year. I also believe #420 might help with the problem this PR is addressing. @igor-kamil feel free to re-open if you think this is still useable.

PR #238, opened 2019-10-09
[item] Refactor color search

rastislav-chynoransky:
# Type of change

# Checklist:

# On prod server
- [ ] Run items:extract-colors command


config/items.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+return [
+    'colors' => [
+        'count' => 6
```
martin-havala: just thinking: should not we store more colours and use reduced number for related search? so we may get higher coverage for some combinations

Dockerfile
```diff
@@ -15,6 +15,9 @@ RUN docker-php-ext-configure gd --with-jpeg-dir=/usr/include/ \
     zip \
     exif
 
+RUN echo "memory_limit=-1" > $PHP_INI_DIR/conf.d/memory-limit.ini
```
igor-kamil: what about `RUN echo "memory_limit=512M" > $PHP_INI_DIR/conf.d/memory-limit.ini`?

resources/views/form/filter/color_widget.blade.php
```diff
@@ -0,0 +1,39 @@
+<div class="color-picker" id="{{ $id }}">
+    <input type="hidden" name="{{ $name }}" v-bind:value="colorSubstring" />
+    <slider-picker v-model="color" @changemouseup="onChange" />
+</div>
+<script src="https://cdn.jsdelivr.net/npm/vue/dist/vue.js"></script>
+<script src="/js/touch-recognition.js"></script>
```
igor-kamil: again I prefer to use `asset()` helper for such a case

PR #258, opened 2020-02-05
fix relationships listed without values in artist detail

igor-kamil:
# Description

fixed relationships listed without values in artist detail

Fixes [WEBUMENIA-818](https://jira.sng.sk/browse/WEBUMENIA-818)

## Type of change

# Checklist:

### DIFF
--- a/app/Authority.php
+++ b/app/Authority.php
@@ -93,7 +93,7 @@ class Authority extends Model implements IndexableModel, TranslatableContract
 
     public function relationships()
     {
-        return $this->belongsToMany(\App\Authority::class, 'authority_relationships', 'authority_id', 'related_authority_id')->withPivot('type');
+        return $this->belongsToMany(\App\Authority::class, 'authority_relationships', 'authority_id', 'related_authority_id')->withPivot('type')->where('authorities.type', '=', 'person');
     }
 
     public function items()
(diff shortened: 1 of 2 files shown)


app/Authority.php
```diff
@@ -93,7 +93,7 @@ public function events()
 
     public function relationships()
     {
-        return $this->belongsToMany(\App\Authority::class, 'authority_relationships', 'authority_id', 'related_authority_id')->withPivot('type');
+        return $this->belongsToMany(\App\Authority::class, 'authority_relationships', 'authority_id', 'related_authority_id')->withPivot('type')->where('authorities.type', '=', 'person');
```
eronisko: Are there any relationships between non-person authorities? Would it make sense to change it to something like this?
```suggestion
        return $this->belongsToMany(\App\Authority::class, 'authority_relationships', 'authority_id', 'related_authority_id')->withPivot('type')->where('authorities.type', $this->type);
```

igor-kamil: yes. there are. 
e.g. Rudolf Pribiš is member of "Generácia 1909" - https://www.webumenia.sk/autor/8110 - we just don't display other then person authorities yet (others are groups / schools / ... )
but other possible fix would be to use [query scope](https://laravel.com/docs/5.8/eloquent#query-scopes) - however it would limit also admin from showing non-person authorities. 
thus I wrote it like this (also I hope it's just temporary.. and in the future that condition would be removed)

PR #266, opened 2020-02-20
Feature/webumenia 978

martin-havala:
# Description

## Type of change

# Checklist:


public/js/webumenia.js
```diff
@@ -148,3 +148,13 @@ var light_style = [
...
 function isMobileSafari() { return navigator.userAgent.match(/(iPod|iPhone|iPad)/); }
+
+
+function copyLinkToClipboard(url) {
+    var $temp = $("<input>");
+    $("body").append($temp);
+    $temp.val(url).select();
+    document.execCommand("copy");
+    $temp.remove();
+    alert("{{ trans('dielo.copied_to_clipboard') }}");
```
igor-kamil: since this is already inside .js file, it wouldn't get translated (literally  alert with `{{ trans('dielo.copied_to_clipboard') }}` would be displayed)
solution might be to pass the alert message as function parameter (or grab it from element that initiated it using data-attribute)

also it would be much nicer to write JS function belonging to the component alongside the component (into `share_buttons_js.blade.php` ... you might get inspired by `searchbar` component). also with

PR #285, opened 2020-03-22
Feature/webumenia 1039

martin-havala:
# Description

## Type of change

# Checklist:


resources/views/frontend/collection/form.blade.php
```diff
@@ -0,0 +1,16 @@
+@php FormRenderer::setTheme($form, 'filter') @endphp
+
+<section class="filters">
+    <div class="container content-section">
+        <div class="expandable">
+            <div class="row">
+                <div class="col-md-4 col-xs-6 bottom-space">
```
igor-kamil: I was just thinking , whether 
```suggestion
                <div class="col-xs-6">
```
 wouldn't look better (now there is this "empty space" instead of the third filter). but let's discuss it once it's on the test visible to everyone. however. please - no `bottom-space` here

resources/views/frontend/collection/form.blade.php
```diff
@@ -0,0 +1,16 @@
...
+                    @formWidget($form['author'], ['attr' => ['class' => 'js-custom-select']])
+                </div>
+                <div class="col-md-4 col-xs-6 bottom-space">
```
igor-kamil: same here
```suggestion
                <div class="col-xs-6">
```

resources/views/frontend/collection/index.blade.php
```diff
@@ -0,0 +1,162 @@
...
+    <div class="col-xs-6 text-right">
+        @formRow($form['sort_by'], ['attr' => ['class' => 'js-dropdown-select']])
+    </div>
+    </div>
+    <div class="kolekcie">
+        @foreach ($collections as $i=>$collection)
+        <div class="row collection">
+            {{-- <div class="col-sm-2 col-xs-4">
+            	<a href="{!! $collection->getUrl() !!}">
+            		<img src="{!! $collection->getHeaderImage() !!}" class="img-responsive pentagon" alt="{!! $collection->name !!}">
```
igor-kamil: I didn't even know we ever had something like this :D

resources/views/frontend/collection/form.blade.php
martin-havala: but we have _bottom-space_ everywhere, also I'd keep stacking to the left, its similar when you change resolution on other pages, inputs ale left-aligned instead of centering

igor-kamil: ok. stacking to the left make sense. but bottom-space was here in catalog view, because there is another row of input fields... but not in this case. it's making unproportional space in the bottom. 
still we fine-tune these details in release branch. technically it's superb.

PR #301, opened 2020-04-02
[harvest] Don't create relations to non-existing entities

rastislav-chynoransky:
# Description

Till now relations were created even if the related entities didn't exist. Even if that might be handy when harvested data come before its related data, generally it's not a good practice. If we want to preserve this behavior I'd suggest creating hidden entities for later completion.

Fixes https://jira.sng.sk/browse/WEBUMENIA-1337

## Type of change

# Checklist:

### DIFF
--- a/app/Harvest/Importers/AbstractImporter.php
+++ b/app/Harvest/Importers/AbstractImporter.php
@@ -126,13 +126,18 @@ abstract class AbstractImporter
         $relation = $model->$field();
         $relatedModelClass = get_class($relation->getRelated());
 
-        $updateIds = [];
+        $updatedIds = [];
 
         foreach ($relatedRows as $relatedRow) {
             $data = $this->mappers[$field]->map($relatedRow);
             $conditions = $this->getConditions($field, $data);
-            $existing = $relatedModelClass::where($conditions)->first();
-            $relatedModel = $existing ?: new $relatedModelClass;
+            $relatedModel = $relatedModelClass::where($conditions)->first();
+
+            if (!$relatedModel && !$createRelated) {
+                continue;
+            }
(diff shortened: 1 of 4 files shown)


tests/Harvest/Importers/ItemImporterTest.php
```diff
@@ -87,6 +88,31 @@ public function testImportAuthorityPivotData() {
         $this->assertEquals('iné/other', $other->pivot->role);
     }
 
+    public function testImportNonExistingAuthority()
+    {
+        // testing for specific bug where records in pivot table
```
eronisko, line 93: Might be worth referencing the JIRA issue here (if it merits the explanation)

app/Harvest/Importers/AbstractImporter.php
```diff
@@ -141,23 +146,18 @@ protected function processBelongsToMany(Model $model, $field, array $relatedRows
...
         }
 
         $relatedKeyName = $relation->getQualifiedRelatedPivotKeyName();
         $relatedKeyName = explode('.', $relatedKeyName);
         $relatedKeyName = end($relatedKeyName);
-        $notUpdated = $relation->whereNotIn($relatedKeyName, $updateIds)->get();
+        $notUpdated = $relation->whereNotIn($relatedKeyName, $updatedIds)->get();
         if (!$notUpdated->isEmpty()) {
             $relation->detach($notUpdated);
         }
```
eronisko, line 163: Could the [`sync`](https://laravel.com/api/5.8/Illuminate/Database/Eloquent/Relations/Concerns/InteractsWithPivotTable.html#method_sync) method be used here?

PR #303, opened 2020-04-13
feature/WEBUMENIA-1336: enable random ordering within catalogue

martin-havala:
# Description

Option to sort items in catalogue randomly

## Type of change

# Checklist:


app/Elasticsearch/Repositories/ItemRepository.php
```diff
@@ -381,6 +382,34 @@ protected function buildHueRangeQuery(float $hue, float $diff): array
...
+
+            $q =  isset($body['query']) ? $body['query'] : ['match_all' => new \stdClass];
+
+            $body['query'] = [
+                'function_score' => [
+                    'query' => $q,
+                    'functions' => [
+                        [
+                            'random_score' => [
+                                'seed' => crc32(Session::getId()),
```
igor-kamil, line 395: nice. this is actually the proper approach - in comparing with the previous implementation, it avoids duplicates on next pages. however - it returns you always the same results for given session (it's missing something like "randomize")
this feature is used mostly by @ustengrav so it might be nice to hear his opinion...
I am just thinking about "randomizing" every time you _start_ to sort by random. 
this might be achieved by generating random seed and storing it into session everytime you are on the first  page  (`$sortBy === 'random'` and `$request->getFrom()==0`)  
what do you think @martin-havala ?

martin-havala: actually I was thinking about new seed on each page refresh, because you need to store it only for 'load more' functionality

igor-kamil: yes. basically you wrote the same as I did above :D just with less words. so would you do it?

martin-havala: I did not :) your's solution would trigger new seed everytime you change sorting, mine would everytime you refresh page using same sorting.. or just lost in translation
i can try

eronisko: FWIW I quite like this simple approach. We could also use the current date as the seed so that you get different results on different days.

But I guess it really depends on who uses this and why.

martin-havala: But @ustengrav would like to get new random set with any filter/sorting change ¯\_(ツ)_/¯

PR #315, opened 2020-05-07
bugfix/WEBUMENIA-1364: fixed article-saving problem

martin-havala:
# Description

fixes article saving problem while keeping ConvertEmptyStringsToNull middleware,
small locale fixes and admin tweaks

[bugfix/WEBUMENIA-1364](https://jira.sng.sk/browse/WEBUMENIA-1364)
[bugfix/WEBUMENIA-1348](https://jira.sng.sk/browse/WEBUMENIA-1348)

## Type of change

# Checklist:

### DIFF
--- a/app/Collection.php
+++ b/app/Collection.php
@@ -93,6 +93,11 @@ class Collection extends Model implements TranslatableContract
         return $result_path;
     }
 
+    public function scopeContentImages($query)
+    {
+        return parseUrls($this->text);
+    }
+
     public function scopePublished($query)
     {
         return $query->where('published_at', '<', Carbon::now());
--- a/app/Collection.php
+++ b/app/Collection.php
@@ -93,7 +93,7 @@ class Collection extends Model implements TranslatableContract
         return $result_path;
     }
 
-    public function scopeContentImages($query)
+    public function getContentImages()
     {
         return parseUrls($this->text);
     }
--- a/app/Helpers/general.php
+++ b/app/Helpers/general.php
@@ -349,3 +349,30 @@ function formatName($name) {
 function starts_with_upper($str) {
     return (bool)preg_match('/^[[:upper:]]/u', $str);
 }
+
+function hasTranslationValue($locale, $attributes)
+{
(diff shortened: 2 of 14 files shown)


app/Collection.php
```diff
@@ -93,6 +93,11 @@ public function getResizedImage($resize)
         return $result_path;
     }
 
+    public function scopeContentImages($query)
```
rastislav-chynoransky, line 96:
```suggestion
    public function getContentImages()
```
Scopes should be used in context of query building

app/Helpers/general.php
```diff
@@ -349,3 +349,30 @@ function formatName($name) {
 function starts_with_upper($str) {
     return (bool)preg_match('/^[[:upper:]]/u', $str);
 }
+
+function hasTranslationValue($locale, $attributes)
```
rastislav-chynoransky, line 353: Sorry, maybe I'm late to discussion... Was there a problem if an empty translation would have been saved?

martin-havala: not really, we'd just wanted to prevent the storing of empty translations

app/Helpers/general.php
```diff
@@ -349,3 +349,30 @@ function formatName($name) {
...
+{
+    foreach ($attributes as $attribute) {
+        if (\Input::get($locale . '.' . $attribute) != null) {
```
rastislav-chynoransky, line 356: Is loose comparison here intentional?

app/Collection.php
igor-kamil: this is better now. let's keep it for now.
but for future - it's also possible to use "custom/virtual attributes"
if you name that function `getContentImagesAttribute()`, you can then access it with simple `$article->content_images` instead of `$article->getContentImages()`
to read more, check the docs -> https://laravel.com/docs/5.8/eloquent-mutators#defining-an-accessor

app/Helpers/general.php
igor-kamil: good point @rastislav-chynoransky

PR #318, opened 2020-05-08
feature/WEBUMENIA-1340: Add authority ID to artwork records;

martin-havala:
# Description

- item->author is now connected to authority, administration is offering multiselect
- it's still possible to add authors as a simple strings
 
- no bulk edit yet

Fixes [WEBUMENIA-1340](https://jira.sng.sk/browse/WEBUMENIA-1340)

## Type of change

# Checklist:


martin-havala:
okay, I :

- left authors as a unified field for both authors and authorities, because it's reused in several places as a source for authors list. 
- set item_form -> author delimiter to ";", so we can add anyone, even out of authorities table in desired format _Surname, Name_. 
- I've set order of Authors within Item.getAuthorsWithLinks, because it was sorted by _authority_id_ (alphabetically)

TODO: I guess  we should completely separate **items.author** from **authorities_item** pivot logic and then update logic used for views accordingly(so either Item.getAuthorsWithLinks will use the same logic as getTitleWithAuthors and getAuthorsAttribute . or vice versa, because one is mixing linked authorities and items.author and rest of those are using items.author only).. or I'm missing some logic/relationship here

PR #323, opened 2020-05-14
[item] Fix indexing work types

rastislav-chynoransky:
Fixes https://jira.sng.sk/browse/WEBUMENIA-1291

## Type of change

# Checklist:

### DIFF
--- a/app/Item.php
+++ b/app/Item.php
@@ -582,6 +582,7 @@ class Item extends Model implements IndexableModel, TranslatableContract
 
     public function getIndexedData($locale)
     {
+        $work_types = $this->makeArray($this["work_type:$locale"], ', ');
         return [
             'id' => $this->id,
             'identifier' => $this->identifier,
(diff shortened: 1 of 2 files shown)


app/Item.php
```diff
@@ -598,7 +596,7 @@ public function getIndexedData($locale)
             'is_free' => $this->isFree(),
             'authority_id' => $this->authorities()->pluck('id'),
             'view_count' => $this->view_count,
-            'work_type' => is_array($work_types) ? reset($work_types) : '',
+            'work_type' => $this->makeArray($this["work_type:$locale"], ', '),
```
igor-kamil, line 599: ok for now. however this doesn't preserve hierarchy. 
also the generated values might looks bit inconsistent: 
Screenshot 2020-05-15 at 20 09 30
("`voľná`" is both for grafika and kresba)
what do you think about trying https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-pathhierarchy-tokenizer.html in the future?

rastislav-chynoransky: I agree. One (easier) option would be to complete the name of deeper work type hierarchy with parent names. The other option with the hierarchical tokenizer you proposed would probably need to display the hierarchy in filter somehow.

mcudrnak:
Oh, i should have thought about this too when discussing it with you. 

- yes, usually we can split the values, like "maliarstvo, koláž" or "úžitkové umenie, sklo, nádoba"
and keep them separate
- but sometimes, you can't do that, at the same level of "hierarchy" you have terms like"maliarstvo, funkčné"  - the order needs should reverted into "funčkné maliarstvo", the comma doesn't imply hierarchy, it's a remainder of an obsolete rule for writing the values in the vocabulary.
- in that case, you wont solve this by path hierarchy tokenizer, because it won't tell you when to separate the value by comma and when to keep it as one (or revert the order)

I think we have to discuss this first before releasing it, or postpone. We should also fix it at the source - edit the values in CEDVU.

PR #349, opened 2020-09-09
[authority] Display organization type

rastislav-chynoransky:
Fixes https://jira.sng.sk/browse/WEBUMENIA-1266

## Type of change

# Checklist:

### DIFF
--- a/app/Authority.php
+++ b/app/Authority.php
@@ -286,9 +286,8 @@ class Authority extends Model implements IndexableModel, TranslatableContract
 
     private static function getNoImage($sex = 'male')
     {
-        $filename = 'no-image' . ($sex ? "-$sex" : '') . '.jpeg';
-
-        return "/images/no-image/autori/$filename";
+        if ($sex) return "/images/no-image/autori/no-image-$sex.jpeg";
+        return "/images/no-image/autori/no-image.jpeg";
     }
 
     public function getIndexedData($locale)
--- a/app/Authority.php
+++ b/app/Authority.php
@@ -322,4 +322,9 @@ class Authority extends Model implements IndexableModel, TranslatableContract
             $this->save();
         }
     }
+
+    public function isCorporateBody()
+    {
+        return $this->type === 'corporate body';
+    }
 }
--- a/app/Authority.php
+++ b/app/Authority.php
@@ -194,39 +194,6 @@ class Authority extends Model implements IndexableModel, TranslatableContract
         return URL::to('autor/'.$authority_id);
     }
 
-    public function getDescription()
-    {
-        $parts = [];
-
-        if ($this->birth_year) {
-            $parts[] = $this->birth_year;
-        }
-        if ($this->birth_year && $this->birth_place) {
-            $parts[] = $this->birth_place;
-        }
-        if ($this->birth_year && $this->death_year) {
-            $parts[] = '-';
-        }
-        if ($this->death_year) {
-            $parts[] = $this->death_year;
-        }
-        if ($this->death_year && $this->death_place) {
-            $parts[] = $this->death_place;
-        }
-
-        $description = implode(' ', $parts);
-
-        if ($this->roles) {
-            $description = sprintf(
-                '%s. Role: %s',
-                $description,
-                implode(', ', $this->roles)
-            );
-        }
-
-        return $description;
-    }
-
     public static function getImagePathForId($id, $has_image, $sex = 'male', $full = false, $resize = false)
     {
         if (!$has_image && !$full) {
--- a/app/Authority.php
+++ b/app/Authority.php
@@ -93,7 +93,7 @@ class Authority extends Model implements IndexableModel, TranslatableContract
 
     public function relationships()
     {
-        return $this->belongsToMany(\App\Authority::class, 'authority_relationships', 'authority_id', 'related_authority_id')->withPivot('type')->where('authorities.type', '=', 'person');
+        return $this->belongsToMany(\App\Authority::class, 'authority_relationships', 'authority_id', 'related_authority_id')->withPivot('type');
     }
 
     public function items()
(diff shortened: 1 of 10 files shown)


app/Authority.php
```diff
@@ -289,7 +286,7 @@ public static function getImagePathForId($id, $has_image, $sex = 'male', $full =
 
     private static function getNoImage($sex = 'male')
     {
-        $filename = 'no-image-'.$sex.'.jpeg';
+        $filename = 'no-image' . ($sex ? "-$sex" : '') . '.jpeg';
 
         return "/images/no-image/autori/$filename";
```
eronisko, line 291:
```suggestion
        if ($sex) return "/images/no-image/autori/no-image-$sex.jpeg";
        return "/images/no-image/autori/no-image.jpeg";
```

app/Authority.php
```diff
@@ -194,40 +194,37 @@ public static function detailUrl($authority_id)
...
+
+        if ($this->birth_year) {
+            $parts[] = $this->birth_year;
+        }
+        if ($this->birth_year && $this->birth_place) {
+            $parts[] = $this->birth_place;
+        }
+        if ($this->birth_year && $this->death_year) {
+            $parts[] = '-';
+        }
```
eronisko, line 209: This is a huge improvement, but since it's somewhat purely presentation-related, could we parse this out from the actual description in the view and just strip off HTML tags? Just a suggestion.

resources/views/autor.blade.php
```diff
@@ -31,7 +31,7 @@
     </section>
 @endif
 
-<section class="author detail content-section" itemscope itemtype="http://schema.org/Person">
+<section class="author detail content-section" itemscope itemtype="http://schema.org/{{ $author->type == 'corporate body' ? 'Organization' : 'Person' }}">
```
eronisko, line 34: I can see these checks for "is corporate body?" a bit. Maybe a helper like `$author->is_corporate_body` (or `is_person`) would be useful?
```suggestion
<section class="author detail content-section" itemscope itemtype="http://schema.org/{{ $author->is_corporate_body ? 'Organization' : 'Person' }}">
```

resources/views/autor.blade.php
```diff
@@ -66,13 +66,46 @@
                         <p class="lead">{{ trans('authority.alternative_names') }} <em>{!! implode("</em>, <em>", $author->formatedNames) !!}</em></p>
                     @endif
                     <p class="lead">
-                        {!! $author->getDescription(true, true) !!}
+                        @if ($author->birth_date || $author->birth_place)
+                            *
+                        @endif
+
+                        @if ($author->birth_date)
+                            <span itemprop="{{ $author->type === 'corporate body' ? 'foundingDate' : 'birthDate' }}">{{ $author->birth_date }}</span>
```
eronisko, line 74: I think it might be simpler to enclose the two options (corporate vs person) in an `@if`, e.g.

```blade
@if($person->is_corporate_body)
 {{-- view logic for corporations --}}
@else
 {{-- view logic for persons--}}
@endif
```

Bit less DRY, but I think easier on the eye without all the inline ifs.

resources/views/autor.blade.php
```diff
@@ -66,13 +66,46 @@
...
+                        @if ($author->birth_place)
+                            <a href="{{ route('frontend.author.index', ['place' => $author->birth_place]) }}" itemprop="{{ $author->type === 'corporate body' ? 'foundingPlace' : 'birthPlace' }}">{{ $author->birth_place }}</a>
+                        @endif
+
+                        @if (($author->birth_date || $author->birth_place) && ($author->death_date || $author->death_place))
+                            &ndash;
+                        @endif
+
+                        @if ($author->death_date || $author->death_place)
+                            &#x271D;
```
eronisko, line 86: Would this work?
```suggestion
                            ✝
```

resources/views/autor.blade.php
```diff
@@ -3,7 +3,7 @@
 @section('og')
 <meta property="og:title" content="{!! $author->formatedName !!}" />
 
-<meta property="og:description" content="{!! $author->getDescription(false, false, true) !!}" />
+<meta property="og:description" content="{{ $author->getDescription() }}" />
```
eronisko, line 6: As I mentioned above, it might be worth trying to encapsulate the description in a component, and then do something like

```suggestion
<meta property="og:description" content="{{ html_strip(@component('components.author_description', $author) }}" />
```
I honestly dont know if it'll work but if it does, we could delete some more code :)

eronisko (commented):
Looks good overall, a big improvement. Just have a few suggestions regarding the view logic.

app/Authority.php
rastislav-chynoransky: I think it's great idea! I usually try to preserve behavior as it is, but description is not a huge deal actually.

resources/views/autor.blade.php
rastislav-chynoransky: I've realized there is so much common logic that I'd prefer not to split it. I've just improved indentation at one critical place. What do you think?

app/Http/Controllers/AuthorController.php
```diff
@@ -38,9 +38,20 @@ public function getDetail($id)
...
         }
 
+        $htmlDescription = view('components.authority_description')
+            ->with('author', $author)
+            ->render();
+
+        $description = strip_tags($htmlDescription);
+        $description = preg_replace('/\s+/', ' ', $description);
+        $description = preg_replace('/ ,/', ',', $description);
+        $description = trim($description);
```
eronisko, line 48: BTW, Rails has these _view helpers_, where you can put presentation code that's logic-heavy. That would work really well here. (But this here is fine of course).
https://www.rubyguides.com/2020/01/rails-helpers/

eronisko (approved):
That reads much better, thanks! ❤

resources/views/autor.blade.php
eronisko: Looks good!

PR #352, opened 2020-09-29
[import] Add PNP trienale importer

rastislav-chynoransky:
Fixes https://jira.sng.sk/projects/PNP/issues/PNP-4

## Type of change

# Checklist:


app/Importers/PnpTrienaleImporter.php
```diff
@@ -0,0 +1,151 @@
...
+            'linoryt' => 'linocut',
+            'serigrafie' => 'serigraphy',
+            'litografie' => 'lithography',
+            'počítačová grafika' => 'computer graphics',
+            'kombinovaná technika' => 'combined technique',
+            'mezzotinta' => 'mezzotint',
+            'mědiryt' => 'chalcography',
+            'šablona na batice' => 'stencil on batik',
+            'vernis mou' => 'vernis mou',
+        ],
```
rastislav-chynoransky, line 80: @mcudrnak please check these translations

app/Item.php
```diff
@@ -122,6 +124,8 @@ class Item extends Model implements IndexableModel, TranslatableContract
         'belongsToManyUpdatedExistingPivot',
     ];
 
+    protected $useTranslationFallback;
```
eronisko, line 127: Is this actually doing anything?

eronisko (changes requested):
Looks good overall, just one thing: I don't think `$useTranslationFallback` is doing anything.

app/Importers/PnpTrienaleImporter.php
```diff
@@ -0,0 +1,151 @@
...
+    {
+        if ($locale !== 'cs') {
+            return null;
+        }
+
+        return [
+            'award' => $record['Ocenění:'],
+            'award_category' => $record['Kategorie:'],
+        ];
+    }
```
eronisko, line 150: Actually, now that I've thought about this, why not add two additonal columns instead? We don't have a native JSON support in the current DB version, so I'd hold off with using JSON like this. 

This way we're losing many of the Eloquent features and we can't really use or update these fields very easily.

I'll happily be persuaded otherwise, but right now, I have a strong preference for using plain-old columns for this. cc @igor-kamil

eronisko (changes requested):

app/Item.php
rastislav-chynoransky: It's unexisting otherwise

app/Importers/PnpTrienaleImporter.php
rastislav-chynoransky: This is the most universal thing we'd come up with. We try to design the structure for future custom data that we don't use (award and award category will be displayed only at PNP website).

app/Item.php
eronisko: But does its existence make any difference? I don't see it being used anywhere.

rastislav-chynoransky: Ah, sure... https://github.com/Astrotomic/laravel-translatable/blob/master/src/Translatable/Translatable.php#L445

rastislav-chynoransky: We want to set it `false` so it wouldn't fallback to SK language

eronisko: Ah, my bad! Sorry and thanks.

app/Importers/PnpTrienaleImporter.php
eronisko: My 2c: it is a good approach when there's a native support for it in the DB and application. Right now, it isn't really much different from saving them as `?award=abc&award_category=xyz` or any other arbitrary format. You have to parse them every time you want to use them and any effort for change (e.g. "migration") would be a search & replace exercise.

I totally get the need and I think this is another good reason to move to MySQL 5.7. Moving these sparse columns to a JSON column shouldn't be too much hassle after that.

igor-kamil (changes requested):
since "Inventární číslo" column is missing in the current XLS file, it ends up with `Undefined index`
please add auto-generating of ID/identifier

igor-kamil (approved):
sorry my fail. it works nicely. ready to go!

PR #356, opened 2020-10-06
[authority] Add matcher

rastislav-chynoransky:
Fixes https://jira.sng.sk/browse/WEBUMENIA-1385

## Type of change

# Checklist:


app/Parsers/AuthorParser.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+namespace App\Parsers;
+
+class AuthorParser
```
eronisko, line 5: Just my opinion: I would prefer to keep this inside AuthorityMatcher as long as there's no other entity reusing it. We already have 20 namespaces under `app/` and it could just be a simple class method instead.

app/Console/Commands/MatchAuthorities.php
```diff
@@ -0,0 +1,66 @@
...
+                            $message = sprintf('Multiple authorities matched (%s)', $multiple);
+                        }
+
+                        fwrite($fp, $message . PHP_EOL);
+                        return null;
+                    })
+                    ->filter();
+
+                $changes = $item->authorities()->syncWithoutDetaching($ids);
+                $item->authorities()->updateExistingPivot($changes['attached'], ['automatically_matched' => true]);
```
eronisko, line 52: Could this be done outside of the 'foreach' loop? It may speed things up a bit since we might be doing this more than once per authority, no?

rastislav-chynoransky: Could you please suggest a change?

rastislav-chynoransky: Also this is more-or-less once run command, so we don't really need to focus on optimizations.

app/Parsers/AuthorParser.php
rastislav-chynoransky: I'm not fan of static methods but yes.

app/Console/Commands/MatchAuthorities.php
eronisko: > Also this is more-or-less once run command, so we don't really need to focus on optimizations.

Fair enough.

eronisko (approved):

PR #357, opened 2020-10-08
Update model status during harvesting of failed

rastislav-chynoransky:
Fixes https://jira.sng.sk/browse/WEBUMENIA-1353

## Type of change

# Checklist:

### DIFF
--- a/app/Harvest/Harvesters/AuthorityHarvester.php
+++ b/app/Harvest/Harvesters/AuthorityHarvester.php
@@ -5,7 +5,7 @@ namespace App\Harvest\Harvesters;
 use App\Authority;
 use App\Harvest\Importers\AuthorityImporter;
 use App\Harvest\Repositories\AuthorityRepository;
-use App\Harvest\Result;
+use App\Harvest\Progress;
 use App\SpiceHarvesterRecord;
 
 class AuthorityHarvester extends AbstractHarvester
@@ -14,7 +14,7 @@ class AuthorityHarvester extends AbstractHarvester
         parent::__construct($repository, $importer);
     }
 
-    protected function harvestSingle(SpiceHarvesterRecord $record, Result $result, array $row = null) {
+    public function harvestRecord(SpiceHarvesterRecord $record, Progress $progress, array $row = null) {
         if ($row === null) {
             $row = $this->repository->getRow($record);
         }
@@ -30,7 +30,7 @@ class AuthorityHarvester extends AbstractHarvester
             }
         }
 
-        return parent::harvestSingle($record, $result, $row);
+        parent::harvestRecord($record, $progress, $row);
     }
 
     // @todo could be moved to mapper if mapper'd have access to whole row
--- a/app/Harvest/Progress.php
+++ b/app/Harvest/Progress.php
@@ -22,6 +22,9 @@ class Progress
     protected $deleted = 0;
 
     /** @var int */
+    protected $failed = 0;
+
+    /** @var int */
     protected $total;
 
     public function __construct(\DateTime $createdAt = null) {
@@ -48,8 +51,16 @@ class Progress
         return $this->deleted;
     }
 
+    public function getFailed() {
+        return $this->failed;
+    }
+
     public function getProcessed() {
-        return $this->inserted + $this->updated + $this->skipped + $this->deleted;
+        return $this->inserted +
+            $this->updated +
+            $this->skipped +
+            $this->deleted +
+            $this->failed;
     }
 
     public function incrementInserted() {
@@ -66,6 +77,10 @@ class Progress
 
     public function incrementDeleted() {
         $this->deleted++;
+    }
+
+    public function incrementFailed() {
+        $this->failed++;
     }
 
     public function getTotal() {
--- a/app/Harvest/Progress.php
+++ b/app/Harvest/Progress.php
@@ -4,7 +4,7 @@ namespace App\Harvest;
 
 use Carbon\Carbon;
 
-class Result
+class Progress
 {
     /** @var \DateTime */
     protected $createdAt;
@@ -20,6 +20,9 @@ class Result
 
     /** @var int */
     protected $deleted = 0;
+
+    /** @var int */
+    protected $total;
 
     public function __construct(\DateTime $createdAt = null) {
         $this->createdAt = $createdAt ?: Carbon::now();
@@ -45,7 +48,7 @@ class Result
         return $this->deleted;
     }
 
-    public function getTotal() {
+    public function getProcessed() {
         return $this->inserted + $this->updated + $this->skipped + $this->deleted;
     }
 
@@ -63,5 +66,13 @@ class Result
 
     public function incrementDeleted() {
         $this->deleted++;
+    }
+
+    public function getTotal() {
+        return $this->total;
+    }
+
+    public function setTotal($total) {
+        $this->total = $total;
     }
 }
--- a/app/Http/Controllers/SpiceHarvesterController.php
+++ b/app/Http/Controllers/SpiceHarvesterController.php
@@ -228,8 +228,7 @@ class SpiceHarvesterController extends Controller
         $all = Input::get('reindex', false);
 
         $this->dispatch(new HarvestJob($harvest, $from, $to, $all));
-
-        return Redirect::route('harvests.index');
+        return Redirect::back();
     }
 
     public function refreshRecord($record_id)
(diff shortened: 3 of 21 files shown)


app/SpiceHarvesterRecord.php
```diff
@@ -32,4 +34,22 @@ public function scopeFailed($query)
...
+    {
+        try {
+            $onProcess();
+
+            $this->failed_at = null;
+            $this->error_message = null;
+        } catch (\Exception $e) {
+            $this->failed_at = Carbon::now();
+            $this->error_message = $e->getMessage();
+            $progress->incrementFailed();
```
eronisko, line 48: Did you mean to have `app('sentry')->captureException($e);` here too?

tests/Harvest/Importers/AuthorityImporterTest.php
```diff
@@ -43,7 +43,7 @@ public function testUpdateRelated() {
         $row = $this->getData();
         $importer = $this->initImporter($row);
 
-        $authority = $importer->import($row, $result = new Result());
+        $authority = $importer->import($row, $result = new Progress());
```
eronisko, line 46: Total nit pick, but these are making me uneasy:
```suggestion
        $authority = $importer->import($row, new Progress());
```

app/SpiceHarvesterHarvest.php
```diff
@@ -63,5 +65,51 @@ public function getStatusClassAttribute()
...
+    public function advance(Progress $progress)
+    {
+        $this->status_messages = trans('harvest.status_messages.progress', [
+            'current' => $progress->getProcessed(),
+            'total' => $progress->getTotal(),
+        ]);
+        $this->save();
+    }
 
+    public function queue()
```
eronisko, line 109: Another nit pick. `queue()` implies a getter to me.

```suggestion
    public function enqueue()
```
```suggestion
    public function addToQueue()
```

app/SpiceHarvesterHarvest.php
```diff
@@ -63,5 +65,51 @@ public function getStatusClassAttribute()
...
+            $this->status = SpiceHarvesterHarvest::STATUS_ERROR;
+            $this->status_messages = trans('harvest.status_messages.error', [
+                'error' => $e->getMessage(),
+            ]);
+        } finally {
+            $this->save();
+        }
+    }
+
+    public function advance(Progress $progress)
```
eronisko, line 100: This indicates a 'move forward' action (when in fact, it just reflects the state of `$progress`)
```suggestion
    public function updateStatusMessages(Progress $progress)
```

app/SpiceHarvesterHarvest.php
```diff
@@ -63,5 +65,51 @@ public function getStatusClassAttribute()
...
+                'deleted' => $progress->getDeleted(),
+                'skipped' => $progress->getSkipped(),
+                'failed' => $progress->getFailed(),
+                'time' => Carbon::now()->diffInSeconds($progress->getCreatedAt()),
+            ]);
+        } catch (\Exception $e) {
+            $this->status = SpiceHarvesterHarvest::STATUS_ERROR;
+            $this->status_messages = trans('harvest.status_messages.error', [
+                'error' => $e->getMessage(),
+            ]);
```
eronisko, line 94: Again, just checking. Did you mean to include Sentry log here?

app/Http/Controllers/SpiceHarvesterController.php
```diff
@@ -246,6 +245,6 @@ public function harvestFailed($id)
         $harvest = SpiceHarvesterHarvest::findOrFail($id);
         $this->dispatch(new HarvestFailedJob($harvest));
         Session::flash('message', 'Harvest failed job has been queued');
-        return Redirect::route('harvests.index');
+        return Redirect::back();
```
eronisko, line 248: `Redirect::route('harvests.index')` seemed a bit safer to me. Do we have a 'flash' message spot for every view this could redirect back to?

app/Harvest/Progress.php
```diff
@@ -64,4 +78,16 @@ public function incrementSkipped() {
...
+        $this->failed++;
+    }
+
+    public function getTotal() {
+        return $this->total;
+    }
+
+    public function setTotal($total) {
+        $this->total = $total;
+    }
```
eronisko, line 92: I don't think this is necessary (but it's just my opinion and I see you have a  consistent convention going on here)

app/Harvest/Progress.php
```diff
@@ -45,8 +51,16 @@ public function getDeleted() {
...
+    public function getFailed() {
+        return $this->failed;
+    }
+
+    public function getProcessed() {
+        return $this->inserted +
+            $this->updated +
+            $this->skipped +
+            $this->deleted +
+            $this->failed;
```
eronisko, line 63: Very clear 👌

app/Harvest/Harvesters/AuthorityHarvester.php
```diff
@@ -30,7 +30,7 @@ protected function harvestSingle(SpiceHarvesterRecord $record, Result $result, a
             }
         }
 
-        return parent::harvestSingle($record, $result, $row);
+        parent::harvestRecord($record, $progress, $row);
```
eronisko, line 33: Just checking: no `return` necessary here?

eronisko (commented):
Quite a re-factor! 💪 I have a few notes, but this is 99% good-to-go.

database/factories/ModelFactory.php
```diff
@@ -88,14 +88,15 @@
 
 $factory->define(\App\SpiceHarvesterHarvest::class, function (Faker\Generator $faker) {
     return [
-        'type' => 'item',
+        'type' => $faker->randomElement(['item', 'authority']),
```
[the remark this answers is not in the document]
rastislav-chynoransky: So do you suggest not filling it in the factory?

database/factories/ModelFactory.php
```diff
@@ -88,14 +88,15 @@
...
         'base_url' => $faker->url,
         'metadata_prefix' => $faker->word,
         'set_spec' => $faker->word,
         'set_name' => $faker->word,
         'set_description' => $faker->word,
         'status_messages' => $faker->sentence,
         'initiated' => $faker->date,
+        'status' => \App\SpiceHarvesterHarvest::STATUS_QUEUED,
```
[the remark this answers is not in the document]
rastislav-chynoransky: You're right, I think we can keep it without status.

app/SpiceHarvesterRecord.php
rastislav-chynoransky: I'm thinking of not sending these exceptions to sentry as we track errors in admin and those are usually data-related.

app/Http/Controllers/SpiceHarvesterController.php
rastislav-chynoransky: I've used it here as a quick fix for keeping user on the same page number.

app/Harvest/Harvesters/AuthorityHarvester.php
rastislav-chynoransky: The `return` value seemed unnecessary (and unused) to me.

app/Harvest/Progress.php
rastislav-chynoransky: Hm, those are actually used
https://github.com/SlovakNationalGallery/webumenia.sk/pull/357/files#diff-4033abadf28e58540dbb44939447fb461cb4d1f08f89684bcd1a7a872e4e5dfaR104
https://github.com/SlovakNationalGallery/webumenia.sk/pull/357/files#diff-6b7e8d26a84896bf71a62c18ffb5cbce4a3d145506ae55b8a65aa049f02ee430R25
https://github.com/SlovakNationalGallery/webumenia.sk/pull/357/files#diff-6b7e8d26a84896bf71a62c18ffb5cbce4a3d145506ae55b8a65aa049f02ee430R51

database/factories/ModelFactory.php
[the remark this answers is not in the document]
eronisko: I'd prefer keeping it as it was -- setting it to `item`

app/Harvest/Progress.php
eronisko: What I meant was using the basic `$progress->total` and `$progress->total = 123` instead of getters/setters. But feel free to ignore this comment, it would be a pointless change.

database/factories/ModelFactory.php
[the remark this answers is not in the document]
rastislav-chynoransky: But the factory is general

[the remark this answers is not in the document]
eronisko: I try to go with 'just enough data' approach, e.g. a factory should return a valid object with the least data possible. I.e. if `SpiceHarvesterHarvest` does not need `type` in order to be usable, then the factory should not define it.

But I think in this case, it does need the `type`, and I think `item` is a good default. If we need to test some `authority`-specific behavior, there's always an option to override it. If you think `authority` is a better default, then that's fine too. 

I'd just avoid flip-flopping randomly between them, because that would make the tests upredictable.

eronisko (approved):
Nice, thanks!

PR #362, opened 2020-10-14
[elasticserach] fix too large result window

rastislav-chynoransky:
Fixes https://jira.sng.sk/browse/WEBUMENIA-1453

## Type of change

# Checklist:

### DIFF
--- a/app/Http/Controllers/AbstractSearchRequestController.php
+++ b/app/Http/Controllers/AbstractSearchRequestController.php
@@ -16,7 +16,7 @@ abstract class AbstractSearchRequestController extends Controller
 {
     use CreatesForms;
 
-    protected $perPage = 18;
+    protected $perPage = 20;
 
     protected $repository;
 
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/AbstractSearchRequestController.php
```diff
@@ -63,6 +63,12 @@ protected function getIndexData()
 
         $response = $this->repository->search($searchRequest);
         $total = $this->repository->count($searchRequest);
+        $max = max(1, ceil($response->getTotal() / $this->perPage));
+
+        if ($currentPage > $max) {
+            return abort(404);
+        }
```
eronisko, line 70: I haven't seen the Sentry issue, but the message looks like https://jira.sng.sk/browse/WEBUMENIA-1418. I'm happy to be proved wrong, but I don't think this'll solve this problem:

* The check for `$currentPage > $max` comes too late -- at this point the search has already been executed (and generated the same error)
* The `$max` (aka `$lastPage`) code is already implemented in [`LengthAwarePaginator`](https://github.com/laravel/framework/blob/78eb4dabcc03e189620c16f436358d41d31ae11f/src/Illuminate/Pagination/LengthAwarePaginator.php#L51), so I don't think the problem is in the paginator.

On a separate note, I think the right thing to do here is to return an empty result set instead of a 404.

eronisko (changes requested):
If I'm right about this being an older issue, I don't think this'll solve it.

Not sure where to look next, though. Are we're setting the search window `size` and `from` correctly?

app/Http/Controllers/AbstractSearchRequestController.php
rastislav-chynoransky: Ah, sorry, this is wrong at so many levels...

PR #376, opened 2020-10-29
WEBUMENIA-1467 Automatically deploy "develop" branch to test

eronisko:
# Description

Successful run: https://github.com/SlovakNationalGallery/webumenia.sk/actions/runs/336393101

## Type of change

# Checklist:

### DIFF
--- a/.github/workflows/develop-autodeploy.yml
+++ b/.github/workflows/develop-autodeploy.yml
@@ -1,0 +1,32 @@
+name: Deploy "develop" branch
+
+on:
+  push:
+    branches:
+      - develop
+
+jobs:
+  deploy:
+    name: Deploy
+    runs-on: ubuntu-18.04
+    steps:
+    - uses: actions/checkout@v2
+    - name: Get Composer Cache Directory
+      id: composer-cache
+      run: echo "::set-output name=dir::$(composer config cache-files-dir)"
+    - name: Restore Composer cache
+      uses: actions/cache@v2
+      with:
+        path: ${{ steps.composer-cache.outputs.dir }}
+        key: ${{ runner.os }}-composer-${{ hashFiles('**/composer.lock') }}
+        restore-keys: |
+          ${{ runner.os }}-composer-
+    - uses: shivammathur/setup-php@v2
+      with:
+        php-version: 7.4
+    - run: composer install --no-progress --no-interaction --no-dev
+    - uses: atymic/deployer-php-action@0.1.1
+      with:
+          ssh-private-key: ${{ secrets.DEPLOY_SSH_PRIVATE_KEY }}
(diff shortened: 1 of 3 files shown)


eronisko:
Ah, looks like this broke some tests, hang on...

eronisko:
Ready now. I've gone down a rabbit hole of `composer update` and had to come up the way I had come :D

.github/workflows/develop-autodeploy.yml
```diff
@@ -0,0 +1,33 @@
...
+    - name: Restore Composer cache
+      uses: actions/cache@v2
+      with:
+        path: ${{ steps.composer-cache.outputs.dir }}
+        key: ${{ runner.os }}-composer-${{ hashFiles('**/composer.lock') }}
+        restore-keys: |
+          ${{ runner.os }}-composer-
+    - uses: shivammathur/setup-php@v2
+      with:
+        php-version: 7.4
```
igor-kamil, line 26: for test.webumenia.sk we use `php7.2` for now. 
(locally I already use php7.4 for webumenia development and it works nicely... so it shouldn't be any issue)
maybe we can update also virtual host to start using `php7.4-fpm` later

igor-kamil (approved):
LGTM

config/elasticsearch/client.php
```diff
@@ -2,6 +2,7 @@
 
 return [
     'hosts' => [
```
rastislav-chynoransky, line 4: Doesn't this need to be an array of hosts?

eronisko: It still is: https://github.com/SlovakNationalGallery/webumenia.sk/blob/e6e6f0f4059f636053c3de7c6c1594e408e5a833/config/elasticsearch/client.php

eronisko: Now I get what you meant :) Well spotted.

config/elasticsearch/client.php
```diff
@@ -2,6 +2,7 @@
...
-        sprintf('http://%s:%d', env('ES_HOST'), env('ES_PORT'))
+        'host' => env('ES_HOST', 'localhost'),
+        'port' => env('ES_PORT', '9200'),
```
rastislav-chynoransky, line 6:
```suggestion
        [
                'host' => env('ES_HOST', 'localhost'),
                'port' => env('ES_PORT', '9200'),
        ]
```
I mean like this. Not sure if that would not be interpreted as two hosts: `localhost`, `9200`.

eronisko: You're totally right. This PR is already merged in, but here's the fix: https://github.com/SlovakNationalGallery/webumenia.sk/pull/380

PR #378, opened 2020-11-02
[filesystem] skip links

rastislav-chynoransky:
# Description

[Links are not supported by flysystem](https://github.com/thephpleague/flysystem/issues/599). We want to symlink images from storage to image server, but it'd fail on searching directory for csv files.

## Type of change

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ All notable changes to this project will be documented in this file[^1].
 ### Fixed
 - wrong method name in harvester
 
+### Changed
+- skip links in flysystem
+
 ## [2.8.0] - 2020-10-26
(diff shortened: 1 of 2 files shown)


igor-kamil (approved):
looks ok

eronisko:
BTW, we're using symlinks with register-architektury and I haven't noticed any issues. Are these some special links, like across hostnames?

rastislav-chynoransky:
> BTW, we're using symlinks with register-architektury and I haven't noticed any issues. Are these some special links, like across hostnames?

when reading directory with flysystem (default laravel file manager). those are links from storage dir. it is possible to refactor importer so it can read outside of storage, but let leave that for later

PR #385, opened 2020-11-09
Release/2.11.0

rastislav-chynoransky:


CHANGELOG.md
```diff
@@ -3,6 +3,17 @@ All notable changes to this project will be documented in this file[^1].
 
 ## [Unreleased]
 
+## [2.11.0] - 2020-11-10
+### Added
+- WEBUMENIA-1467 Automatically deploy "develop" branch to test
```
igor-kamil, line 8: do we want to use task names in changelog? 
I don't mind... Just asking 😉 
I was thinking, since these tasks are outside of github issues - it wouldn't be that much meaningful for outsiders. however for insiders - us - it's useful... I admit. 
@eronisko, @rastislav-chynoransky  ?

igor-kamil (approved):
LGTM

CHANGELOG.md
rastislav-chynoransky: I've been thinking of removing them just to keep things simple and consistent.

eronisko: It would be also good to ascertain why we need a Changelog in the first place, because to me it is not 100% clear. We don't have any consumers that I'm aware of and not having it reference commits/PRs/issues kinda makes it redundant IMHO.

PR #404, opened 2020-11-27
Feature/refactor authors with links

rastislav-chynoransky:
## Type of change

# Checklist:


routes/web.php
```diff
@@ -265,8 +266,24 @@ function()
             }
         }
 
+        $authorsWithoutAuthority = app(AuthorityMatcher::class)
+            ->matchAll($item, $onlyExisting = true)
+            ->filter(function (\Illuminate\Support\Collection $authorities) use ($item) {
+                return $authorities->isEmpty();
+            })
+            ->keys();
```
eronisko, line 274: This kinda feels like it could be encapsulated inside AuthorityMatcher or even Item.

routes/web.php
```diff
@@ -265,8 +266,24 @@ function()
...
+
+        $htmlAuthorities = view('components.item_authorities')
+            ->with('item', $item)
+            ->with('authors', $authorsWithoutAuthority)
+            ->render();
+
+        $htmlAuthorities = trim($htmlAuthorities);
+        $htmlAuthorities = trim($htmlAuthorities, ',');
```
eronisko, line 282: Could we have an `$authorities` variable or `$item->matched_authorities` and pass that to the views? The `$htmlAuthorities` name seems kinda redundant if we don't have any other `$authorities`. (But note I don't fully understand what AuthorityMatcher does)

Also all of this seems like view (i.e. not necessarily controller) business, could it live in blade instead?

app/Item.php
```diff
@@ -523,29 +523,6 @@ public function scopeRelated($query, Item $item, $locale = null)
             ->orderBy('related_work_order');
     }
 
-    public function getAuthorsWithLinks()
```
eronisko, line 526: Nice 👍

eronisko:
BTW, thanks for making this a separate PR! 💐

resources/views/components/item_authorities.blade.php
```diff
@@ -0,0 +1,16 @@
+@foreach($item->authorities as $authority)
+    @if ($authority->pivot->role === 'autor/author')
+        <span itemprop="creator" itemscope itemtype="http://schema.org/Person">
+            <a class="underline" href="{{ $authority->getUrl() }}" itemprop="sameAs">
+                <span itemprop="name">{{ formatName($authority->name) }}</span><!--
+            --></a><!--
+        --></span>,
```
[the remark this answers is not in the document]
rastislav-chynoransky: I did this to split it to multiple lines and to get rid of whitespace before a comma.

routes/web.php
rastislav-chynoransky: I can imagine creating new method `matchAllExisting` instead of passing an argument, but to keep it reusable in other places, I'd keep filtering outside.

routes/web.php
rastislav-chynoransky: The trick is that I wanted to trim a trailing comma here. Also I think that the matching logic is a bit heavy for a view.

routes/web.php
rastislav-chynoransky: But yes, we could move this to `Item` instead.

resources/views/components/item_authorities.blade.php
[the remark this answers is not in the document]
eronisko: Ah ok. It wasn't obvious to me at all, could you maybe add it to the comment?

routes/web.php
eronisko: Would `$loop->last` help? https://laravel.com/docs/8.x/blade#the-loop-variable

Apart from that it's a pretty simple component include IMHO.

rastislav-chynoransky: You'd need to repeat the logic three times.

eronisko: Or how about making a single authority -- instead of the full list-- the component? Then we could do all the comma business in an outer loop.

rastislav-chynoransky: Ok, I'll try what it'd do with the whitespace.

resources/views/dielo.blade.php
```diff
@@ -52,7 +52,9 @@
                 <div class="col-md-10 col-md-offset-1 text-center content-section">
                     <h1 class="nadpis-dielo" itemprop="name">{!! $item->title !!}</h1>
                     <h2 class="inline">
-                        {!! implode(', ', $item->getAuthorsWithLinks()) !!}
+                        @foreach($item->authorities->toBase()->merge($item->getAuthorsWithoutAuthority()) as $author)
```
eronisko, line 55: It would be great to do this in a single `$item->...` call, but that's just a nit pick.

eronisko (approved):
Looking good! 🚀 Thanks for working in the changes.

PR #409, opened 2020-12-03
WEBUMENIA-1501 Basic state management

eronisko:
This defines a simple and a 'kitchen sink' component showing some of the store's actions and getters.

Note this **will** generate a bunch of warnings in the console and break the color and year picker (https://jira.sng.sk/browse/WEBUMENIA-1508). The component should still work nevertheless.


igor-kamil (approved):
👏  very nice! thank you. also local storage is probably the best option for current scenario.

PR #412, opened 2020-12-04
WEBUMENIA-1503 User collection page

eronisko:
* Basic 'my collection' view
* Moved paging link into a component
* 'clear all' button

Labels are not translated (since I don't think we're at that stage yet)

The way this currently works is that the 'my collection' link in navbar is pointing to a URL based on items currently in the storage (ids are added as parameters). It's kinda dodgy but it means we don't have to do Ajax calls for the initial load. @igor-kamil let me know what you think :)

### DIFF
--- a/app/Http/Controllers/UserCollectionController.php
+++ b/app/Http/Controllers/UserCollectionController.php
@@ -1,0 +1,16 @@
+<?php
+
+namespace App\Http\Controllers;
+
+use App\Item;
+use Illuminate\Http\Request;
+
+class UserCollectionController extends Controller
+{
+
+    public function show(Request $request)
+    {
+        $items = Item::whereIn('id', $request->input('ids', []))->simplePaginate(24);
+        return view('frontend.user-collection.show', compact('items'));
+    }
+}
--- a/resources/js/app.js
+++ b/resources/js/app.js
@@ -26,6 +26,7 @@ require('./components/clipboard-button')
 window.Vue = require('vue')
 Vue.component('user-collections-examples', require('./components/user-collections/Examples.vue').default);
 Vue.component('user-collections-nav-link', require('./components/user-collections/NavLink.vue').default);
+Vue.component('user-collections-clear-button', require('./components/user-collections/ClearButton.vue').default);
 Vue.component('color-widget', require('./components/ColorWidget.vue').default);
 Vue.component('year-slider', require('./components/YearSlider.vue').default);
 
--- a/resources/js/app.js
+++ b/resources/js/app.js
@@ -25,6 +25,7 @@ require('./components/clipboard-button')
 // Vue components
 window.Vue = require('vue')
 Vue.component('user-collections-examples', require('./components/user-collections/Examples.vue').default);
+Vue.component('user-collections-nav-link', require('./components/user-collections/NavLink.vue').default);
 Vue.component('color-widget', require('./components/ColorWidget.vue').default);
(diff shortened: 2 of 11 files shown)


app/Http/Controllers/UserCollectionController.php
```diff
@@ -0,0 +1,16 @@
...
+
+use App\Item;
+use Illuminate\Http\Request;
+
+class UserCollectionController extends Controller
+{
+
+    public function show(Request $request)
+    {
+        $items = Item::whereIn('id', $request->input('ids', []))->simplePaginate(24);
```
igor-kamil, line 13: really just a nitpick. and I know it looks like this everywhere else in the code... but what about writing it cleaner - without specifying explicit number? 
I have no _proper solution_ from the top of my head and it can be also nice standalone "cleaning task" in the future.
just some hints:
- define it as controller property `private $per_page = 24;`
- set it in global config `config('app.per_page')` - with eventual overiding with custom settings (e.g. `$request->input('per_page', config('app.per_page'))`)
- use constant

resources/js/app.js
```diff
@@ -25,6 +25,8 @@ require('./components/clipboard-button')
 // Vue components
 window.Vue = require('vue')
 Vue.component('user-collections-examples', require('./components/user-collections/Examples.vue').default);
+Vue.component('user-collections-nav-link', require('./components/user-collections/NavLink.vue').default);
```
igor-kamil, line 28: wow. you are already ahead! ;)

resources/views/frontend/catalog/index.blade.php
```diff
@@ -81,21 +81,7 @@
...
-                                    </text>
-                                </svg>
-                            </a>
-                            <div class="page-load-status" style="display: none">
-                                <p class="infinite-scroll-request" id="infscr-loading"><i class="fa fa-refresh fa-spin fa-lg"></i></p>
-                                <p class="infinite-scroll-last text-muted">{{ utrans('katalog.catalog_finished') }}</p>
-                                <p class="infinite-scroll-error">{{ utrans('katalog.catalog_finished') }}</p>
-                            </div>
-                        @endif
+                        @include('components.load_more', ['paginator' => $paginator, 'isotopeContainerSelector' => '#iso'])
```
igor-kamil, line 84: 👏

resources/views/components/load_more.blade.php
```diff
@@ -0,0 +1,46 @@
...
+</div>
+@endif
+
+@section('javascript')
+@parent
+<script>
+    $(document).ready(function(){
+        var $container = $('{{ $isotopeContainerSelector }}');
+
+        $container.infiniteScroll({
```
igor-kamil, line 23: I didn't realize it until this CR. but when there is not enough results for pagination (<24) - there is a warning in the console: `Bad InfiniteScroll path option. Next link not found: #next  Disabling InfiniteScroll`
would be nice to cover this case - while making the component for `load_more` (otherwise I am happy to see it! 😄  )

igor-kamil (changes requested):
everything looks smoothly. please just check the issue with initialisation of infinite scroll when there is no pagination

resources/views/components/load_more.blade.php
eronisko: Will be fixed in https://jira.sng.sk/browse/WEBUMENIA-1512

igor-kamil (approved):

PR #414, opened 2020-12-07
use asset_timed for og:image

igor-kamil:
# Description

This is a fix to overcome facebook caching of `og:image` ( https://stackoverflow.com/questions/7572441/facebook-like-showing-cached-version-ogimage-way-to-refresh-or-reindex-it ) - to enable new "sharing preview" after og:image is changed.

Re-using already written method `asset_timed()` to include timestamp in image path (e.g. `<meta property="og:image" content="https://www.webumenia.sk/images/clanky/psicek-a-macicka-sa-tesia.jpg?1433999652" />`)

Fixes [WEBUMENIA-1513](https://jira.sng.sk/browse/WEBUMENIA-1513)

## Type of change

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 All notable changes to this project will be documented in this file[^1].
 
 ## [Unreleased]
+## [2.13.1] - 2020-12-07
+### Fixed
+- Use method asset_timed to avoid caching of og:image
+
 ## [2.13.0] - 2020-11-27
 ### Added
 - Add 'copy citation' button to artwork detail
--- a/app/Helpers/general.php
+++ b/app/Helpers/general.php
@@ -133,7 +133,7 @@ function asset_timed($path, $secure = null)
     if (file_exists($file)) {
         return asset($path, $secure) . '?' . filemtime($file);
     } else {
-        throw new \Exception('The file "'.$path.'" cannot be found in the public folder');
+        return asset($path, $secure); // fallback to default asset() to avoid exception
     }
 }
 
--- a/resources/views/autor.blade.php
+++ b/resources/views/autor.blade.php
@@ -5,7 +5,7 @@
 <meta property="og:description" content="{{ $description }}" />
 <meta property="og:type" content="object" />
 <meta property="og:url" content="{!! Request::url() !!}" />
-<meta property="og:image" content="{!! URL::to( $author->getImagePath() ) !!}" />
+<meta property="og:image" content="{!! asset_timed( $author->getImagePath() ) !!}" />
 <meta property="og:site_name" content="Web umenia" />
 @stop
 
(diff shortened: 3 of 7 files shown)


eronisko (commented):
I'm not sure whether this would actually work. Have you tried it? See https://stackoverflow.com/a/31022974/1167094:

> We recently found that Facebook was caching URLs using a query string against the relative URL and that the query string was being ignored which messed up a few dynamic images we were serving purely based on the query string.

My money would be on the approach suggested in the answer

eronisko:
Is manually re-caching the page via https://developers.facebook.com/tools/debug not feasible?

igor-kamil:
thank you for the review @eronisko 
no. re-caching via https://developers.facebook.com/tools/debug didn't help. I tried also _batch invalidator_ ... no help. 
It reloads all other data. But still cache the iamge (since the file name didn't change at all):
Screenshot 2020-12-08 at 10 47 09

we already used `?v=xyz` hack for `og:image` in our [1989.sng.sk](https://1989.sng.sk/) project ... when we faced the same issue. and it helped. 
I tested adding the suffix  and it works:
Screenshot 2020-12-08 at 10 58 27

If you don't see any other issue - I would be happy to merge it.

eronisko:
I can now see how this would work, but to be honest I'm still hesitating, because now we'd have to hit the disk every time we want to render a item/author/article page, which is not great for performance.

How about generating the `?v=` using current timestamp? (say, rounded to the nearest hour so we can allow some caching at least).

As an aside, I think the problem stems from the image being named the same as the ID of the collection. This can probably trip up browser caches too, not just the Facebook scraper. Could we perhaps generate a more specific file URL?

igor-kamil:
You are completely right @eronisko - and I admit it shouldn't be permanent fix.
However - in the current world we even calculate the width/height for every artwork image (e.g in catalogue 18 times on every load).  
I decided to check some numbers: https://stackoverflow.com/questions/5850180/cost-of-file-modification-time-checks
And also here is quite valuable discussion: https://stackoverflow.com/questions/56588850/php-filemtime-vs-mysql-last-updated-timestamp-for-image-caching-in-cms
My proposal is to use it for now (because @ustengrav needs it now and the cost of performance/added_value is still reasonable from my perspective). 
And then use the proper solution. And I am up for randomised filename (what we already do for articles)... or maybe we will come up with better idea?
To be honest - calculated timestamp sounds weird - but is not that bad idea (actually we can "fine-tune the caching by hand ...  1-12hours). And mostly - we talk here about THEIR caching. Not OURS. So I don't really mind. It is bit weird. But should be bullet proof. What is your opinion @eronisko ?

eronisko (approved):
Yeah, I understand that this would not be the worst thing that's going on during render (by far). But I still don't think it would be a step in the right direction.

I'd rather use something like `Carbon::now()->startOfDay()`, maybe even right in the blade templates.

All that said, the code looks OK, so if you decide to merge it as it is (and we create a JIRA task for a proper fix), I'd be fine with it.

igor-kamil:
done: https://jira.sng.sk/browse/WEBUMENIA-1516

PR #417, opened 2020-12-11
WEBUMENIA-1505 Basic user collection nav-link animation

eronisko:
I've added a couple of Vue transitions using Animate.css. 

Feel free to use/adjust or even discard them, I just wanted to see how they work :)


igor-kamil (approved):
nice! I like it

PR #433, opened 2021-01-07
return back is-detail=true

igor-kamil:
maybe keep `is-detail=true` for now


eronisko (approved):
Ah. This is why it didn't work: https://stackoverflow.com/a/44602796/1167094

igor-kamil:
thank you @eronisko also for the explanation. however I also prefer this `is-detail=true` for keeping it really explicit :)
/from my point of view/

PR #435, opened 2021-01-08
WEBUMENIA-1523 Editable notice

eronisko:
image

## Type of change

# Checklist:


igor-kamil (approved):
wow! this was done quickly! and it's very nice! thanx!
for a moment I was thinking about, how to make it more universal for any other possible notices.. (we have whole table for only one record ... ) / or keeping history
also in the future - we will need `published_from` and `published_until` (e.g. for christmas alerts)
but for now - this is already good to go!

PR #439, opened 2021-01-14
WEBUMENIA-1532 speed up reindexing of Items

eronisko:
Test (140k items) gets re-indexed in ~20 minutes. This can be probably improved but it might be fine for now...

Should I do authorities too?

@rastislav-chynoransky I've had to make some changes to ItemObserver. Could you please check if they look OK?

## Type of change

# Checklist:

### DIFF
--- a/app/Observers/ItemObserver.php
+++ b/app/Observers/ItemObserver.php
@@ -39,8 +39,8 @@ class ItemObserver
 
     public function belongsToManyAttached($relation, Item $item, $ids)
     {
-        if ($item->exists) {
-            $this->itemRepository->indexAllLocales($item);
+        if ($item->fresh()->exists) {
+            $this->itemRepository->indexAllLocales($item->fresh());
             if ($relation === 'authorities') {
                 Authority::findMany($ids)->each(function (Authority $authority) {
                     $this->authorityRepository->indexAllLocales($authority);
@@ -52,7 +52,7 @@ class ItemObserver
     public function belongsToManyDetached($relation, Item $item, $ids)
     {
         if ($item->exists) {
-            $this->itemRepository->indexAllLocales($item);
+            $this->itemRepository->indexAllLocales($item->fresh());
             if ($relation === 'authorities') {
                 Authority::findMany($ids)->each(function (Authority $authority) {
                     $this->authorityRepository->indexAllLocales($authority);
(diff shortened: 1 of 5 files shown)


app/Observers/ItemObserver.php
```diff
@@ -39,8 +39,8 @@ public function deleting(Item $item)
 
     public function belongsToManyAttached($relation, Item $item, $ids)
     {
-        if ($item->exists) {
-            $this->itemRepository->indexAllLocales($item);
+        if ($item->fresh()->exists) {
```
eronisko, line 42: Note to self: this shouldn't be here.

igor-kamil: maybe dumb question - but why is this necessary?

database/migrations/2021_01_14_134218_add_index_to_item_translations.php
```diff
@@ -0,0 +1,32 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('item_translations', function (Blueprint $table) {
+            $table->index(['item_id']);
```
igor-kamil, line 17: 👍

igor-kamil (approved):
wow! that's a really big leap 🚀 
works nicely

app/Observers/ItemObserver.php
eronisko: Not this one in particular, but the one later (`$this->itemRepository->indexAllLocales($item->fresh());`) is necessary because somehow the observer did not get the latest instance. Not sure whether that's a bug in the events library or whether I misunderstood how it works. @rastislav-chynoransky do you have any idea?

rastislav-chynoransky: @eronisko @igor-kamil Sorry I'm late to the party... Here (https://laravel.com/docs/8.x/eloquent-relationships#the-save-method) the documentation says:

> The save and saveMany methods will not add the new models to any in-memory relationships that are already loaded onto the parent model. If you plan on accessing the relationship after using the save or saveMany methods, you may wish to use the refresh method to reload the model and its relationships

It's basically the same for attaching/detaching related models. So it's better now since we're using stored relationships like tags, images etc. when getting index data.
Thanks!

eronisko: Thanks for the confirmation @rastislav-chynoransky

PR #440, opened 2021-01-14
Do not specify default test branch in deploy config

eronisko:
To enable easier deployments from dev.

I keep modifying the deploy.php config locally.

## Type of change

# Checklist:

### DIFF
--- a/config/deploy.php
+++ b/config/deploy.php
@@ -160,7 +160,6 @@ return [
             'hostname' => 'webumenia.sk',
             'deploy_path' => '/var/www/test.webumenia.sk',
             'user' => 'lab_sng',
-            'branch' => 'develop'
         ],
     ],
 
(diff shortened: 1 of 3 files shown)


config/deploy.php
```diff
@@ -160,7 +160,6 @@
             'hostname' => 'webumenia.sk',
             'deploy_path' => '/var/www/test.webumenia.sk',
             'user' => 'lab_sng',
-            'branch' => 'develop'
```
igor-kamil, line 163: > I keep modifying the deploy.php config locally.

same here 😆 
just to make it clear:
what happens, if you don't specify any `--branch`? 
well [now I see](https://github.com/lorisleiva/laravel-deployer/blob/main/docs/all-options.md) - it will be `master` (not the current branch, right?)

I am up for this. maybe just ping everyone afterwards to clarify how to deploy custom branch

igor-kamil (approved):

config/deploy.php
eronisko: Actually, it deployed my current branch when I tested it, so I'm not sure. I'll test and let everyone know.

PR #442, opened 2021-01-15
WEBUMENIA-1491 fix zoom on related artworks in non-SK locale

eronisko:
Note that this currently matches based on current or local fallback. I think that's a brittle way of matching items to artworks. Not sure why it was done this way, but it might worth re-farctoring and connecting them properly via pivot tables.

## Type of change

# Checklist:


app/Item.php
```diff
@@ -529,10 +529,10 @@ public function scopeHasImage($query, $hasImage = true)
         return $query->where('has_image', '=', $hasImage);
     }
 
-    public function scopeRelated($query, Item $item, $locale = null)
```
eronisko, line 532: This is actually wrong, let me revert this.

igor-kamil (approved):
👌

PR #449, opened 2021-02-12
WEBUMENIA-1543 Authority "sources" links

eronisko:
## Type of change

- [] Bug fix (non-breaking change which fixes an issue)
- [] New feature (non-breaking change which adds functionality)

# Checklist:


database/migrations/2021_02_12_083807_add_type_to_links.php
```diff
@@ -0,0 +1,35 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('links', function (Blueprint $table) {
+            $table->string('type')->after('label')->nullable(false)->index();
```
igor-kamil, line 18: to be honest - here I am just in doubts - whether not to set it to `->default('external')` ... 
but this is just a note. not a request. any reason?

resources/js/components/admin/LinksInput.vue
```diff
@@ -0,0 +1,49 @@
+<template>
```
igor-kamil, line 1: I like to see this component! 👍

igor-kamil (changes requested):

database/migrations/2021_02_12_083807_add_type_to_links.php
eronisko: A link without a `type` is not valid anymore. This way you have to be explicit about what type of link you're creating.

On DB side I prefer things to fail fast rather than silently accept incomplete input -- especially if the default is buried in a migration file.

app/Http/Controllers/AuthorityController.php
```diff
@@ -127,71 +128,45 @@ public function edit($id)
...
+        ));
+
+        $input = convertEmptyStringsToNull(Input::all());
+
+        $authority = Authority::findOrFail($id);
+        $authority->fill($input);
+        $authority->save();
+
+        $links = [
+            ...collect($input['externalLinks'])->map(function($link) {
```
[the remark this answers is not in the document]
eronisko: Great catch, thank you.

app/Http/Controllers/AuthorityController.php
```diff
@@ -127,71 +128,45 @@ public function edit($id)
...
+            Authority::$rules, 
+            [
+                'externalLinks.*.url' => Link::$rules['url'],
+                'externalLinks.*.label' => Link::$rules['label'],
+                'sourceLinks.*.url' => Link::$rules['url'],
+                'sourceLinks.*.label' => Link::$rules['label'],
+            ]
+        ));
+
+        $input = convertEmptyStringsToNull(Input::all());
```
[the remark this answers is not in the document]
eronisko: I thought that middleware was there but I wasn't sure wheter we ended up using it. I'll get rid of this and see what happens :)

eronisko:
@igor-kamil thanks for the thorough review! I've made the changes you requested.

igor-kamil (approved):
👌

PR #452, opened 2021-02-18
Feature/webumenia 1547 final touches

igor-kamil:
# Description

fix reported issues - https://jira.sng.sk/browse/WEBUMENIA-1547

## Type of change

# Checklist:

### DIFF
--- a/resources/less/content.less
+++ b/resources/less/content.less
@@ -88,13 +88,13 @@
         padding-top: 2 * @padding-large-vertical;
         padding-bottom: @padding-large-vertical;
     }
-    a {
-        color: @gray-dark !important;
-        border-bottom: @grey solid 2px !important;
+    a:not(.slick-slide)  {
+        color: @gray-dark;
+        border-bottom: @gray-light solid 2px;
         &:hover,
         &:active {
-            color: @gray-dark !important;
-            border-bottom: @gray-dark solid 2px !important;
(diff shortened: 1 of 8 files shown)


resources/less/content.less
```diff
@@ -70,31 +55,18 @@
...
 .collection {
-    &.content-header,
-    &.content-section {
-        padding-top: 2 * @padding-large-vertical;
-        padding-bottom: @padding-large-vertical;
-    }
-    a {
-        color: @gray-dark !important;
-        border-bottom: @grey solid 2px !important;
+    a:not(.slick-slide):not(.readmore-js-toggle)  {
```
eronisko, line 63: This may be a bit of a gotcha if you don't know where to look... Would it make sense to create a separate rule for these two cases instead? Even if it means duplicating some.

eronisko (approved):

resources/less/content.less
igor-kamil: I see your point. Original idea was to automaticaly underline every link in artwork description / collection text / article ...  but side-effect is, that it's also underlining images in embeded carousels + readmore toggle button in long artwork descriptions
This is the fix to ignore them. 
And right now I don't see it nicer to define rules for `.slick-slide` and `. readmore-js-toggle` just to "undo" these properties.

PR #469, opened 2021-03-18
escape attributes in artwork detail

igor-kamil:
# Description

Bit out of the scope of the bug report... but to avoid future bug reports... 

Fixes [WEBUMENIA-1548](https://jira.sng.sk/browse/WEBUMENIA-1548)

## Type of change

# Checklist:

### DIFF
--- a/app/Item.php
+++ b/app/Item.php
@@ -319,7 +319,7 @@ class Item extends Model implements IndexableModel, TranslatableContract
         foreach ($authors_array as $author) {
             $authors[$author] = formatName($author);
         }
-        
+
         return $authors;
     }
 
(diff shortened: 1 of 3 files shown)


app/Item.php
```diff
@@ -565,7 +565,7 @@ public function getAuthorsWithLinks()
 
     public function getTitleWithAuthors($html = false)
     {
-        $dash = ($html) ? ' &ndash; ' : ' - ';
+        $dash = ($html) ? ' – ' : ' - ';
         return implode(', ', $this->authors)  . $dash .  $this->title;
```
eronisko, line 569:
```suggestion
        return implode(', ', $this->authors)  . ' – ' .  $this->title;
```

I'd guess we don't even need the `$html` argument in this method?

eronisko (approved):
I ♥ this change!

`getTitleWithAuthors` could be slimmed down (see my comment) but this is good to go nevertheless.

PR #475, opened 2021-03-25
Laravel 6: Composer updates

eronisko:
Please note that 

* this is goes into a separate branch so I won't be making changes to CHANGELOG until we go merge into develop
* the code at this revision is _not_ functional. I just want to keep PRs smaller, so I'll be splitting them up


igor-kamil (approved):
👌 good to keep it in separated branch for now

PR #476, opened 2021-03-25
Laravel 6: Entrust replacement

eronisko:
Replacing https://github.com/Zizaco/entrust with Laravel Gates https://laravel.com/docs/6.x/authorization#gates

* user `role` is now defined as a column in users table
* Entrust tables (permission, role, etc) are only re-named (not deleted) for now. The migration would be difficult to rollback otherwise & I didn't want to delete them just in case. They'll be deleted in a separate migration.


app/User.php
```diff
@@ -2,24 +2,15 @@
...
-
-    use EntrustUserTrait;
-
-    /**
-     * The database table used by the model.
-     *
-     * @var string
-     */
-    protected $table = 'users';
+    public static $roles = ['admin', 'editor'];
```
igor-kamil, line 10: 'admin' or 'administer'? 
To be honest, I am in doubts, what is better. Whether "old" admin or new "administer" (what probably makes sensw for "role" as "ability").

config/cache.php
```diff
@@ -13,7 +13,7 @@
     |
     */
 
-    'default' => env('CACHE_DRIVER', 'array'), //temp fix for entrust - https://github.com/Zizaco/entrust/issues/541 - should use redis or memcached
+    'default' => env('CACHE_DRIVER', 'file'),
```
igor-kamil, line 16: Temp for long long time. Finaly fixed! ;)

igor-kamil (approved):
I confirm. This gate looks much better, then entrust. Looks fine

app/User.php
eronisko: 'admin' is a role (noun), 'administrate' is an weirdly sounding verb -- the ability of an admin. I totally get why it looks off but I thought 'administrate' would also be easier to search for if we do some refactoring later.

PR #477, opened 2021-03-26
Laravel 6: Replace Input facade with Request

eronisko:
According to https://laravel.com/docs/6.x/upgrade#the-input-facade

* `Input::get` -> `Request::input`
* `Input::<rest>` -> `Request::<rest>`

### DIFF
--- a/resources/views/frontend/catalog/index.blade.php
+++ b/resources/views/frontend/catalog/index.blade.php
@@ -63,7 +63,7 @@
                     </div>
                     <div class="col-sm-12 text-center">
 
-                        {!! $paginator->appends(@Input::except('page'))->render() !!}
+                        {!! $paginator->appends(Request::except('page'))->render() !!}
                         @include('components.load_more', ['paginator' => $paginator, 'isotopeContainerSelector' => '#iso'])
                     </div>
                 </div>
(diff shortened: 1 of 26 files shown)


resources/views/frontend/catalog/index.blade.php
```diff
@@ -63,7 +63,7 @@
                     </div>
                     <div class="col-sm-12 text-center">
 
-                        {!! $paginator->appends(@Input::except('page'))->render() !!}
```
rastislav-chynoransky, line 66: do we know the reason for using `@` here? and why it's not needed anymore?

eronisko: My guess is that it was a typo, couldn't find it among any Blade directives.

rastislav-chynoransky: In PHP it's the error suppression operator

eronisko: 😮 
image

eronisko: Thanks, I didn't know that. I'll double check.

eronisko: I didn't find where this could error out (here's the source code: https://github.com/laravel/framework/blob/6.x/src/Illuminate/Http/Concerns/InteractsWithInput.php#L265) but I've decided to just put them back to prevent introducing any new bugs. Good catch, @rastislav-chynoransky.

rastislav-chynoransky (approved):

PR #489, opened 2021-04-09
do not overwrite authority biography via OAI-PMH

igor-kamil:
# Description

Fixes [WEBUMENIA-1580](https://jira.sng.sk/browse/WEBUMENIA-1580)

## Type of change

# Checklist:


app/Harvest/Mappers/AuthorityMapper.php
```diff
@@ -38,6 +38,8 @@ public function mapSex(array $row) {
         }, $row['sex']);
     }
 
+    // Do not store vp:Biography_Text from OAI in Biography on Web (WEBUMENIA-1580)
+    /*
     public function mapBiography(array $row) {
```
igor-kamil, line 43: I decided to keep it commented out... something back in my head is saying, that it might return... (e.g. I think it's better to store at least "something" - if we have nothing). But definitely right now it's making just troubles. 
Are you OK with that @eronisko ?

eronisko: Well, you know what I'm going to say :D It's dead code, so I'd rather remove it. By the time we need it again, it might be already obsolete. If you delete, it'll keep on living in git, and in our hearts.

eronisko (approved):
image
Yeah!

PR #490, opened 2021-04-09
fix counting clicks on carousel slides

igor-kamil:
# Description

it got apparently lost when working on redesign (and rewriting intro carousel from flickity to slick.js)

## Type of change

# Checklist:

### DIFF
--- a/resources/views/intro.blade.php
+++ b/resources/views/intro.blade.php
@@ -114,6 +114,15 @@ $(document).ready(function(){
         centerPadding: '5vw',
         appendDots: $('.slick-pagination')[0]
     });
+
+    $('.webumeniaCarousel').on('click', 'a.outer-box', function( event ) {
+        event.preventDefault();
+        var id = $(this).data('id');
+        var url = $(this).attr('href');
+        $.get('/slideClicked', {'id': id}).done(function( data ) {
+            window.location.href = url;
+        });
+    });
 });
 
 </script>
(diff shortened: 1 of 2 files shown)


resources/views/intro.blade.php
```diff
@@ -114,6 +114,15 @@
...
         appendDots: $('.slick-pagination')[0]
     });
+
+    $('.webumeniaCarousel').on('click', 'a.outer-box', function( event ) {
+        event.preventDefault();
+        var id = $(this).data('id');
+        var url = $(this).attr('href');
+        $.get('/slideClicked', {'id': id}).done(function( data ) {
+            window.location.href = url;
+        });
```
eronisko, line 124: Would this work if we didn't do the `preventDefault()`? Something like 
```suggestion
        var id = $(this).data('id');
        $.get('/slideClicked', {'id': id})
```

igor-kamil: I tested it and it works in my case (and my browser). but I am unsure - how would it behave in other browsers. 
any reason for this suggestion? cleaner code? or that you don't care for the ajax response (so you don't need to wait unti you receive the response.. also even when it fails it will keep links working). what do you think @ajobi ? should it work also this way smoothly everywhere?

eronisko: > also even when it fails it will keep links working

Exactly this. We'd be promoting the tracking action as the primary action and the redirect only happens if that request succeeds. Not to mention the delay this adds.

Btw, probably best approach is to use GA for this kind of tracking. We could add "ref=carousel" to the link or be even more sophisticated:
https://support.google.com/tagmanager/answer/6106716?hl=en

eronisko (approved):

PR #502, opened 2021-04-22
WEBUMENIA-1603 do not parse authority links when harvesting from CEDVU

eronisko:
## Type of change

# Checklist:


rastislav-chynoransky:
@eronisko is this just a quick fix? should we be merging those?

igor-kamil:
@eronisko I think @rastislav-chynoransky is asking, whether this is the proper/final fix? 
this is removing some functionality. just to fix some specific issue. 
it's ok as a quick fix to just ignore these links
but otherwise I would appreciate to investigate it further. 
do we want to ignore them? or just append them (and keep existing links)?

rastislav-chynoransky (approved):

eronisko:
@igor-kamil I understand you concern. Just to clarify, this is fixing an issue _and_ removing functionality that is no longer used as a result.

PR #503, opened 2021-04-27
Fix/webumenia 1610 harvest

igor-kamil:
# Description

```
- Harvester error logging
- Harvester missing total in progress
```

Fixes [WEBUMENIA-1610](https://jira.sng.sk/browse/WEBUMENIA-1610)

## Type of change

# Checklist:

### DIFF
--- a/app/Harvest/Repositories/AbstractRepository.php
+++ b/app/Harvest/Repositories/AbstractRepository.php
@@ -49,8 +49,7 @@ abstract class AbstractRepository
                 $row = $this->getDataRecursively($record, $this->fieldMap);
                 yield $row[0];
             }
-        } catch (MalformedResponseException $e) {
-        }
+        } catch (MalformedResponseException $e) { }
     }
 
     /**
@@ -62,7 +61,12 @@ abstract class AbstractRepository
     public function getTotal(SpiceHarvesterHarvest $harvest, \DateTime $from = null, \DateTime $to = null) {
         $endpoint = $this->endpointFactory->createEndpoint($harvest);
         $records = $endpoint->listRecords($harvest->metadata_prefix, $from, $to, $harvest->set_spec);
-        return $records->getTotalRecordCount();
+
+        $total = 0;
+        try {
+            $total = $records->getTotalRecordCount();
+        } catch (MalformedResponseException $e) { }
+        return $total;
     }
 
     /**
--- a/app/Harvest/Repositories/AbstractRepository.php
+++ b/app/Harvest/Repositories/AbstractRepository.php
@@ -45,15 +45,24 @@ abstract class AbstractRepository
         $records = $endpoint->listRecords($harvest->metadata_prefix, $from, $to, $harvest->set_spec);
 
         try {
-            $total = $records->getTotalRecordCount();
-
             foreach ($records as $record) {
                 $row = $this->getDataRecursively($record, $this->fieldMap);
                 yield $row[0];
             }
         } catch (MalformedResponseException $e) {
-            $total = 0;
         }
+    }
+
+    /**
+     * @param SpiceHarvesterHarvest $harvest
+     * @param \DateTime $from
+     * @param \DateTime $to
+     * @return integer
+     */
+    public function getTotal(SpiceHarvesterHarvest $harvest, \DateTime $from = null, \DateTime $to = null) {
+        $endpoint = $this->endpointFactory->createEndpoint($harvest);
+        $records = $endpoint->listRecords($harvest->metadata_prefix, $from, $to, $harvest->set_spec);
+        return $records->getTotalRecordCount();
     }
 
     /**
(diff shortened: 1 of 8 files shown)


app/Harvest/Repositories/AbstractRepository.php
```diff
@@ -45,17 +45,26 @@ public function getRows(SpiceHarvesterHarvest $harvest, \DateTime $from = null,
...
 
         try {
-            $total = $records->getTotalRecordCount();
-
             foreach ($records as $record) {
                 $row = $this->getDataRecursively($record, $this->fieldMap);
                 yield $row[0];
             }
         } catch (MalformedResponseException $e) {
-            $total = 0;
```
eronisko, line 55: Is a silent failure really the right thing to do here? My preference would be not to catch the exception here, but I don't know this code well enough.

igor-kamil: to be honest. I don't know. I don't even know, why there is this try/catch block... I think it's fair to raise exception, if it doesn't work - @rastislav-chynoransky can you explain?

rastislav-chynoransky: i think if there is an empty result set you can't call getTotalRecordCount on it

rastislav-chynoransky: so it's an exception that should be caught in the new getTotal method, since we don't consider it being an error state that we present to the user

rastislav-chynoransky: but getting data from the oai provider is quite slow already now, i'd suggest not splitting it into two successive calls but to return an object containing `Generator` and total count in it

igor-kamil: it's like @rastislav-chynoransky wrote. this try-catch block was there to handle empty result. it will raise exception both for `getDataRecursively()` and `getTotalRecordCount()` without this try-catch - thus I applied for both.
returning empty set is absolutely OK and common scenario... no reason to store error. 
now it ends up nicely:
```
Spracovaných záznamov: 0
0 nových záznamov
0 bolo upravených
0 bolo zmazaných
0 bolo preskočených
0 zlyhalo.
```

eronisko: As an outsider, it seems strange to me to raise an exception for calling `getTotalRecordCount()` on a empty resultset. I'd expect `0` or `null` (provided empty resulset ~= `[]`). What's the reason for it?

rastislav-chynoransky: That's not in our codebase.
https://github.com/caseyamcl/phpoaipmh/blob/933b4beb1c5d0de675d553ad6f9e545e633ba73b/src/RecordIterator.php#L229

eronisko (changes requested):
Sorry, but I can't approve this yet. It now silently ignores an exception in two places instead of one. I understand that it is a strange decision on the part of the library to return an exception after calling "getTotal()" but I'm convinced that we can prevent this from happening. I know it seems petty to block such a minor PR but silent failures like this can easily add hours of work when debugging.

I'd be more than happy to help figure out a solution. Can you or @rastislav-chynoransky help me understand how this type of failure occurs? Is there any way we could validate the endpoint before starting the iteration?

rastislav-chynoransky:
I remember checking it back then and I couldn't find a way with this lib. According to docs there's no need to catch it. Is it possible our responses are not valid against the oai pmh specs? Could @igor-kamil check? 
https://github.com/caseyamcl/phpoaipmh/tree/master#retrieving-records

igor-kamil:
@rastislav-chynoransky is right
actually - it happened multiple times - in the past when we started to use `caseyamcl/phpoaipmh` , because our oai-pmh provider is custom made solution by EEA - and we found multiple bugs / not following the specification ... since the beginning of using it
@eronisko to simulate it. just launch it for very short time span - e.g. edit in database `completed` to the future  

e.g. when I request http://stary.webumenia.sk/oai-pmh?verb=ListRecords&metadataPrefix=oai_dc&from=2021-04-30T15%3A37%3A42Z&set=GUSvolne - it receives empty records (`<ListRecords/>`) - as expected
running XML validator against for the given scheme fails on: `The Content Of Element 'ListRecords' Is Not Complete. One Of '{"http://www.openarchives.org/OAI/2.0/":record}' Is Expected., Line '4', Column '19'.`

rastislav-chynoransky:
So the provider should return `noRecordsMatch` error message instead (according to docs):
http://www.openarchives.org/OAI/openarchivesprotocol.html#ErrorConditions
We might either try to ask them to implement it (@mcudrnak?) or silence the exception on our side.

eronisko (commented):
I've tried a few things out and I'm still kinda stumped. Being compliant to the OAI standard and returning the error would probably be the way to go, but updating a stabilized API like this one might break things elsewhere.

As a quick-fix I'm proposing narrowing down the exception to its message. It would be nice to not do this in two places (here I'd prefer being DRY, since it's a non-standard logic), but I think that would call for a larger re-factor.

I've also got a work-in-progress version of a solution that @rastislav-chynoransky proposed (returning total and data in one object). I'll submit it as a separate PR, but it's not finished.

igor-kamil:
@eronisko @rastislav-chynoransky I think we are done with this... right?

PR #505, opened 2021-05-03
getAll() returns object with total and data

eronisko:
Proposal based on solution proposed in https://github.com/SlovakNationalGallery/webumenia.sk/pull/503#discussion_r621966737 

TODO: Status is not set to 'complete' if an empty dataset is returned (failing tests).

@igor-kamil feel free to close this if you'd prefer to just get a quick-fix out of the door.

### DIFF
--- a/app/Harvest/Repositories/AbstractRepository.php
+++ b/app/Harvest/Repositories/AbstractRepository.php
@@ -5,6 +5,8 @@ namespace App\Harvest\Repositories;
 use App\Harvest\Factories\EndpointFactory;
 use App\SpiceHarvesterHarvest;
 use App\SpiceHarvesterRecord;
+use Illuminate\Support\LazyCollection;
+use Illuminate\Support\Str;
 use Phpoaipmh\Exception\MalformedResponseException;
 
 abstract class AbstractRepository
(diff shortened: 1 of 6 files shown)


tests/Harvest/Harvesters/AuthorityHarvesterTest.php
```diff
@@ -37,7 +37,7 @@ public function testTryHarvestNoRows()
         ]);
 
         $repositoryMock->expects($this->once())
-            ->method('getRows')
+            ->method('getAll')
             ->willReturn([]);
```
rastislav-chynoransky, line 41:
```suggestion
            ->willReturn((object) [
                'total' => 0,
                'data' => []
            ]);
```

tests/Harvest/Harvesters/ItemHarvesterTest.php
```diff
@@ -65,7 +65,7 @@ public function testHarvestNoRows()
         ]);
 
         $repositoryMock->expects($this->once())
-            ->method('getRows')
+            ->method('getAll')
             ->willReturn([]);
```
rastislav-chynoransky, line 69:
```suggestion
            ->willReturn((object) [
                'total' => 0,
                'data' => []
            ]);
```

rastislav-chynoransky (approved):
🙏

eronisko:
Oooh @rastislav-chynoransky thanks!

app/Harvest/Repositories/AbstractRepository.php
```diff
@@ -38,35 +40,33 @@ public function __construct(EndpointFactory $endpointFactory) {
...
 
         try {
+            $total = $records->getTotalRecordCount();
+        } catch (MalformedResponseException $e) { 
+            // Our OAI API does not return `noRecordsMatch` error code, so handle this particular error
+            if (Str::startsWith("Expected XML element list 'record' missing for verb 'ListRecords'", $e->getMessage())) {
+                $total = 0;
+            }
+
+            throw $e;
```
igor-kamil, line 56:
```suggestion
            } else {
                throw $e;
            }
```

igor-kamil: I would still like to avoid the exception in this case... 
It's completely OK to return 0 updated/new records. This will also prevent `completed` from being saved and kept almost all the harvests most of the time in "error" state. What would probably leads to lower sensitivity - if there is real exception. 
What do you think? 

Screenshot 2021-05-10 at 14 42 00

igor-kamil (commented):
I like it a lot and I think this is much better now. Please see my comment.

app/Harvest/Repositories/AbstractRepository.php
rastislav-chynoransky: good catch! i didn't realise this

PR #512, opened 2021-05-20
EDU: Add vzdelavanie education to main navigation

eronisko:
Fixes WEBUMENIA-1612

### DIFF
--- a/app/Http/Controllers/EducationalArticleController.php
+++ b/app/Http/Controllers/EducationalArticleController.php
@@ -85,6 +85,11 @@ class EducationalArticleController extends Controller
         ));
     }
 
+    public function show(Article $article)
+    {
+        return view('frontend.articles.show', compact('article'));
+    }
+
     private function buildSelectOptions(Collection $optionsWithcounts, $selectedValue = null)
     {
         return $optionsWithcounts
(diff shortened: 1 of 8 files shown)


app/Http/Controllers/EducationalArticleController.php
```diff
@@ -85,6 +85,11 @@ public function index(Request $request)
         ));
     }
 
+    public function show(Article $article)
+    {
+        return view('frontend.articles.show', compact('article'));
```
igor-kamil, line 90: do we want to do it this way? this will duplicate URLs for the same article. the only benefit is, that we will stay in "edu" section
maybe in makes sense. but then we definitely must implement "canonical url":  https://developers.google.com/search/docs/advanced/crawling/consolidate-duplicate-urls

we already this that for artworks -> https://github.com/SlovakNationalGallery/webumenia.sk/blob/develop/resources/views/dielo.blade.php#L23-L25

this can be solved in separated task

igor-kamil (approved):
works fine. please see my comment. feel free to fix that either within this PR or in the  standalone task.

app/Http/Controllers/EducationalArticleController.php
eronisko: Good point

PR #516, opened 2021-05-21
EDU: Display tags and categories in the article preview

eronisko:
image


igor-kamil (approved):
works nicely. I think this should be shown also in article detail (can be done later).

PR #541, opened 2021-08-20
SharedUserCollections: persistence

eronisko:
Migrations, controller, tests

### DIFF
--- a/app/SharedUserCollection.php
+++ b/app/SharedUserCollection.php
@@ -1,0 +1,29 @@
+<?php
+
+namespace App;
+
+use Illuminate\Database\Eloquent\Casts\AsCollection;
+use Illuminate\Database\Eloquent\Model;
+use Illuminate\Support\Str;
+
+class SharedUserCollection extends Model
+{
+    protected $casts = [
+        'items' => AsCollection::class
+    ];
+
+    protected $fillable = [
+        'name',
+        'author',
+        'description',
+        'items',
+    ];
+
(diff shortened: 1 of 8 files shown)


app/SharedUserCollection.php
```diff
@@ -0,0 +1,29 @@
...
+        'name',
+        'author',
+        'description',
+        'items',
+    ];
+
+    protected static function booted()
+    {
+        static::creating(function ($collection) {
+            $collection->public_id = $collection->public_id ?: Str::random(6);
```
igor-kamil, line 25: no check for unique?

resources/views/frontend/shared-user-collections/show.blade.php
```diff
@@ -0,0 +1,51 @@
+@extends('layouts.master') {{-- TODO figure out layout --}}
+
+@section('title')
+    Zdieľanie kolekcie {{-- TODO i18n --}}
```
igor-kamil, line 4: this is quick to implement....

igor-kamil (commented):
overall looks nice. please define translation strings before merge.

app/SharedUserCollection.php
eronisko: There's an index on the DB side. I've figured the probability of a clash is very low (it's a random string of 6 alphanumeric characters in both cases)

igor-kamil (approved):
since this is just part of an epic, it's good to go as it is

PR #545, opened 2021-09-09
Do not set (most) writable_dirs in test

eronisko:
Because lab_sng does not have permissions to `chmod` files owned by data-www

# Description

## Type of change

# Checklist:

### DIFF
--- a/config/deploy.php
+++ b/config/deploy.php
@@ -168,8 +168,6 @@ return [
             'writable_chmod_mode' => '775',
             'writable_dirs' => [
                 'bootstrap/cache',
-                'storage/app/public',
-                'storage/logs',
             ],
         ],
     ],


config/deploy.php
```diff
@@ -168,8 +168,6 @@
             'writable_chmod_mode' => '775',
             'writable_dirs' => [
                 'bootstrap/cache',
-                'storage/app/public',
-                'storage/logs',
```
igor-kamil, line 172: why? 
btw... shouldn't be `'storage'` enough?

eronisko: Because `www-data` user creates (and owns) files within these folders. So `lab_sng` does not have permissions to chmod here.

igor-kamil (approved):
🆗

PR #549, opened 2021-09-24
Webumenia 1668 favorites 2nd design iteration

eronisko:
image
image

### DIFF
--- a/resources/less/components/inline-input.less
+++ b/resources/less/components/inline-input.less
@@ -49,7 +49,7 @@
     text-align: center;
     top: -1rem;
     right: -3rem;
-    color: white;
+    color: @light;
     background-color: @dark;
     padding-top: .4rem;
     font-size: 1.5rem;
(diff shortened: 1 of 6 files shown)


resources/less/components/inline-input.less
```diff
@@ -39,6 +39,25 @@
...
+    display: none;
+  }
+
+  &:hover > &__pencil {
+    display: block;
+    position: absolute;
+    text-align: center;
+    top: -1rem;
+    right: -3rem;
+    color: white;
```
igor-kamil, line 52: [dust]
```suggestion
    color: @light;
```

resources/views/components/user-collections/share-form.blade.php
```diff
@@ -12,7 +12,7 @@
                 name="name" 
                 placeholder="Nazvi svoj výber" {{-- TODO i18n --}}
```
igor-kamil, line 13: [dust] I prefer to see translation strings. 
I trust there will come `i18n` task once :D
but if you used translation strings now - somebody else could take care about that (just to fill in EN/CS versions)

igor-kamil (approved):
looks nice

PR #572, opened 2021-11-18
Newsletter Signup Form Bottom modal

eronisko:
* Adds spatie/laravel-googletagmanager
* Tracks interactions (modal open, dismissed, signed-up) in GTM
* Remembers if modal is dismissed (won't show it for 30 days)
* Remembers if user is signed up won't show the modal OR the inline form.

## Type of change

# Checklist:


resources/views/livewire/newsletter-signup-form-bottom-modal.blade.php
```diff
@@ -0,0 +1,25 @@
...
+        >
+            <div class="row py-5">
+                <div class="visible-lg-block col-lg-1 col-lg-offset-2 text-right pl-0 pt-4">
+                    <svg viewBox="0 0 101 110" fill="none" xmlns="http://www.w3.org/2000/svg">
+                        <rect x="3" y="3" width="94.7368" height="104" stroke="black" stroke-width="6"/>
+                        <rect x="21.5" y="20" width="40" height="40" fill="black"/>
+                        <rect x="47.8652" y="72.7769" width="20" height="20" transform="rotate(-20 47.8652 72.7769)" fill="black"/>
+                    </svg>
+                </div>
+                <div class="col-lg-7 pr-0 pl-2">
```
igor-kamil, line 19:
```suggestion
                <div class="col-lg-7">
```

otherwise it breaks the spacing on mobile
Screenshot 2021-11-22 at 00 44 53
:

app/Http/Livewire/NewsletterSignupFormBottomModal.php
```diff
@@ -0,0 +1,34 @@
...
+             Date::now()->toIso8601String(),
+             Date::now()->addDays(30)->diffInMinutes()
+        );
+        $this->emit('trackNewsletterSignup', 'modalDismissed');
+    }
+
+    public function render()
+    {
+        if (Cookie::has('newsletterSubscribedAt')) return '';
+        if (Cookie::has('newsletterSignupModalDismissedAt')) return '';
```
igor-kamil, line 31: any reason to have it here?
wouldn't it be better to use condition to decide, whether to even show this component?

igor-kamil (approved):
please see my note regarding the spacing on mobile
there no other blocking issues
it looks very nice!

config/googletagmanager.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+return [
+
+    /*
+     * The Google Tag Manager id, should be a code that looks something like "gtm-xxxx".
+     */
+    'id' => 'GTM-W8KZ265',
```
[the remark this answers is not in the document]
eronisko: Thanks. To me, it's more practical to hard-code this sort of thing when in reality, we don't use any different value. (It used to be included in master layout template anyway, so this keeps it the same). Swapping it out for an env var would be still easy if that requirement arises.

resources/views/layouts/master.blade.php
```diff
@@ -37,15 +37,9 @@
...
-		<script>(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
-		new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
-		j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
-		'https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);
-		})(window,document,'script','dataLayer','GTM-W8KZ265');</script>
-		<!-- End Google Tag Manager -->
+		@include('googletagmanager::head')
 
+		@if (App::environment() == 'production')
 		<script>
```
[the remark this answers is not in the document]
eronisko: I have looked into this in the past. I don't recall why I didn't do it in the end, but I think there was a catch.

resources/views/layouts/master.blade.php
```diff
@@ -108,10 +97,10 @@
 	    </a>
 	</div>
 
+	@livewireScripts
```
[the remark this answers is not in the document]
eronisko: Yes. It's for the laravel-vue-lang to work.

resources/views/livewire/newsletter-signup-form-bottom-modal.blade.php
eronisko: Good catch!

app/Http/Livewire/NewsletterSignupFormBottomModal.php
eronisko: I find this pattern quite common in the 'component' world. The component itself has an insight into whether it should show or not. It is this component that sets these cookies, so it makes sense for the component to evaluate them. The 'inline' component doesn't work this way (which might be confusing), but its show/hide logic is much simpler.

It's obviously just my preference... but this way the conditions don't have to pollute templates and it elminates some duplication.

config/googletagmanager.php
```diff
@@ -0,0 +1,27 @@
...
+
+    /*
+     * Enable or disable script rendering. Useful for local development.
+     */
+    'enabled' => env('APP_ENV') == 'production',
```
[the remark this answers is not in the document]
eronisko: I usually just set it to `true` and then not commit it... How about something this?
```suggestion
    'enabled' => env('GOOGLE_TAG_MANAGER_ENABLED', env('APP_ENV') == 'production')
```

app/Http/Livewire/NewsletterSignupFormBottomModal.php
igor-kamil: ok. I agree.

resources/views/livewire/newsletter-signup-form-bottom-modal.blade.php
eronisko: Thank you again. Fixed: 
image

PR #585, opened 2021-12-11
update gitignore

rastislav-chynoransky:
- add `/public/fonts`, `/public/resources`
- remove unused `data/`
- order alphabetically


.gitignore
```diff
@@ -13,17 +13,18 @@ public/sitemap.xml
...
 /public/hot
 /public/js/*LICENSE.txt
 /public/js/app.js
 /public/js/admin.js
 /public/js/vendor.js
 /public/js/manifest.js
-/public/css/admin.css
-/public/css/style.css
-/public/vendor/laravel-filemanager
+/public/resources
```
eronisko, line 25: This is not right. Probably this line is no longer working: https://github.com/SlovakNationalGallery/webumenia.sk/blob/b0e2d107e2957c51a3ed841b70ce79df03b37ab8/webpack.mix.js#L46

It should generate the `default.css` file inside `resources/views/... ` and not `public/resources/views/...` and then it would be covered by line 29/30 in this file. 

It's probably a bug that I caused, sorry.

.gitignore
```diff
@@ -13,17 +13,18 @@ public/sitemap.xml
 .env
 .env.testing
 .phpunit.result.cache
-data/
+/public/css/admin.css
+/public/css/style.css
+/public/fonts
```
eronisko, line 18: This one is also weird. I guess we should probably also remove what's currently in `/public/fonts`: https://github.com/SlovakNationalGallery/webumenia.sk/tree/b0e2d107e2957c51a3ed841b70ce79df03b37ab8/public/fonts and make it 1:1 with `/resources/fonts`... Or remove `resources/...` and keep `public/fonts`? cc @igor-kamil

igor-kamil: ouch this is a mess. yes, I think it should be all together within `/resources/fonts`

.gitignore
eronisko: I believe I've added a fix for this :)

webpack.mix.js
```diff
@@ -43,7 +43,12 @@ mix
     // CSS
     .less('resources/less/admin.less', 'public/css')
     .less('resources/less/style.less', 'public/css')
-    .less("resources/less/mail/default.less", "../resources/views/vendor/mail/html/themes")
+
+    // E-mail themes are expected in resources/ but Mix does not save outside of public/
+    // https://github.com/laravel-mix/laravel-mix/issues/1228
```
igor-kamil, line 48: 👀

.gitignore
igor-kamil: just copy `/public/fonts/slick.*` into `/resources/fonts` and it would be done. the other fonts are already there.

igor-kamil:
I just found font slick was actualy never usef for carousel dots (instead the fallback was taken - GTWalsheim)
Which one do we want to use? 
slick font:
Screenshot 2021-12-20 at 15 40 44

or GTWalsheim:
Screenshot 2021-12-20 at 15 41 50


in case we don't want to use slick font - I am happy to remove it from repository.

cc @ustengrav @mcudrnak

eronisko (approved):
Deployed this to test, LGTM.

igor-kamil (approved):

PR #596, opened 2022-01-03
Feature/webumenia 1584 add gender filter

igor-kamil:
Screenshot 2022-01-03 at 23 05 18

⚠️  This requires to reindex authorities: `php artisan es:reindex authorities`
❓ I am wondering, whether to store translated or non-translated version in the ES index. On one side - it makes sense to store in SK index only SK values. On the other hand it will behave differently when you query DB and ES -> and this inconsistency makes me bit nervous. Also in this way it's harder to appropriate no-image placeholders by gender (what doesn't work anyway right now). What's your opinion @eronisko @rastislav-chynoransky ? 

## Type of change

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 All notable changes to this project will be documented in this file[^1].
 
 ## [Unreleased]
+### Added
+- filter by gender to authorities
 
 ## [2.40.0] - 2021-12-23
 ### Added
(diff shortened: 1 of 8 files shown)


eronisko:
I would prefer the index to be 1:1 with the DB and translations in Blade/all

igor-kamil:
> I would prefer the index to be 1:1 with the DB and translations in Blade/app

closing in favor of #598

PR #598, opened 2022-01-06
WEBUMENIA-1584 add gender filter for authors + fix no-image

igor-kamil:
the idea behind this is to allow using translations when listing data from elastic search - and keeping the data in es consistent with DB
+
this allow to use `getNoImage` method as before
+ 
finally updated no-image for dead/alive authors from Brano Matis

Screenshot 2022-01-07 at 18 11 51

Screenshot 2022-01-03 at 23 05 18


app/Filter/Forms/Types/AuthoritySearchRequestType.php
```diff
@@ -71,6 +71,7 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                 ->add('nationality', ChoiceType::class, $getChoiceOptions('nationality'))
                 ->add('place', ChoiceType::class, $getChoiceOptions('place'))
                 ->add('sex', ChoiceType::class, $getChoiceOptions('sex'));
+                // ->add('sex', ChoiceType::class, $getChoiceOptions('sex', 'authority.sex')); // @TODO: use this after fix
```
igor-kamil, line 74: this doesn't work as expected. 
when passing translation domain as parameter e.g. `$getChoiceOptions('sex', 'authority.sex')` I end up with `{"error":{"root_cause":[{"type":"index_not_found_exception","reason":"no such index [webumenia_authorities_authority.sex]"` - apparently it uses `$translation_domain` as a `$locale` .. 
but I didn't find a workaround for now. any hints @rastislav-chynoransky ?

app/Filter/Forms/Types/AuthoritySearchRequestType.php
```diff
@@ -46,8 +46,8 @@ public function buildForm(FormBuilderInterface $builder, array $options)
             $searchRequest = $event->getData();
             $form = $event->getForm();
 
-            $getChoiceOptions = function ($attribute) use ($searchRequest) {
-                $choices = $this->authorityRepository->listValues($attribute, $searchRequest);
+            $getChoiceOptions = function ($attribute, $translation_domain = null) use ($searchRequest) {
+                $choices = $this->authorityRepository->listValues($attribute, $searchRequest, null, $translation_domain);
```
eronisko, line 50: This should probably be in camelCase (as other vars are)
```suggestion
            $getChoiceOptions = function ($attribute, $translationDomain = null) use ($searchRequest) {
                $choices = $this->authorityRepository->listValues($attribute, $searchRequest, null, $translationDomain);
```

But I'm actually confused by the term "translation domain" entirely :D Is it not the same as "locale" or "lang"?

```suggestion
            $getChoiceOptions = function ($attribute, $locale = null) use ($searchRequest) {
                $choices = $this->authorityRepository->listValues($attribute, $searchRequest, null, $locale);
```

public/images/no-image/autori/no-image-female-dead.svg
```diff
@@ -0,0 +1 @@
...Z"/><path class="cls-2" d="M224.52,124.16a6.31,6.31,0,0,0-1.5-2.23l-6-8.83a3.63,3.63,0,0,0-6.73,2.7l1.79,10.49a6.43,6.43,0,1,0,12.39-2.13"/><path class="cls-2" d="M251,188.33a3.63,3.63,0,0,0-5.66-4.53l-8.25,6.73a6.43,6.43,0,1,0,9.17,7.32Z"/><path class="cls-2" d="M196,72a7.68,7.68,0,1,1-7.67-7.68A7.67,7.67,0,0,1,196,72"/><path class="cls-10" d="M196,72a7.68,7.68,0,1,1-7.67-7.68A7.67,7.67,0,0,1,196,72Z"/><path class="cls-2" d="M199.6,52.49a7.2,7.2,0,0,0-9.83,2.63,7.39,7.39,0,0,0-.83,2.25h0a7.57,7.57,0,0,0-.13,1.57h-1a7.19,7.19,0,1,0-11.9,5.22h0a7.2,7.2,0,0,0,1.31.89l-.49.84a7.83,7.83,0,0,0-1.43-.68h0a7.35,7.35,0,0,0-2.36-.39,7.2,7.2,0,0,0,0,14.39,12.89,12.89,0,0,0,2.36-.4l5.32-3.47.62-.3a7.81,7.81,0,0,1-.62-3A7.65,7.65,0,0,1,195,68.27l.28-.14,4.15-3.07a7.19,7.19,0,0,0,.19-12.57"/><path class="cls-9" d="M199.6,52.49a7.2,7.2,0,0,0-9.83,2.63,7.39,7.39,0,0,0-.83,2.25h0c-.1.52-.38,1.31-.37,1.57h-.49a10,10,0,0,0-1.2-3.83,7.2,7.2,0,1,0-10.94,9.05h0c.41.34.95,1,1.18,1.1l-.24.42c-.22-.14-1-.29-1.55-.47h0a7.35,7.35,0,0,0-2.36-.39,7.2,7.2,0,0,0,0,14.39,12.89,12.89,0,0,0,2.36-.4l5.32-3.47.62-.3a7.81,7.81,0,0,1-.62-3A7.65,7.65,0,0,1,195,68.27l.28-.14,4.15-3.07a7.19,7.19,0,0,0,.19-12.57Z"/></svg>
```
eronisko, line 1: Are these new SVGs optimized?

app/Authority.php
```diff
@@ -205,10 +205,10 @@ public static function detailUrl($authority_id)
         return URL::to('autor/'.$authority_id);
     }
 
-    public static function getImagePathForId($id, $has_image, $sex = 'male', $full = false, $resize = false)
+    public static function getImagePathForId($id, $has_image, $sex = 'male', $death_year = null, $full = false, $resize = false)
```
eronisko, line 208: Why is this even a static function? 😱

app/Authority.php
```diff
@@ -255,17 +255,19 @@ public static function getImagePathForId($id, $has_image, $sex = 'male', $full =
...
     }
 
-    private static function getNoImage($sex = 'male')
+    private static function getNoImage($sex = 'male', $death_year = null)
     {
-        if ($sex) return "/images/no-image/autori/no-image-$sex.jpeg";
-        return "/images/no-image/autori/no-image.jpeg";
+        $filename = "no-image";
+        if ($sex) $filename .= "-$sex";
+        if ($death_year!==null) $filename .= "-dead";
```
eronisko, line 269:
```suggestion
        if ($death_year !== null) $filename .= "-dead";
```

public/images/no-image/autori/no-image-female-dead.svg
igor-kamil: fixed :)

app/Authority.php
igor-kamil: I was asking myself the same question 😄 . Also searching, whether it's being used statically - and it's not. 
It's following the same naming as in `Item` model. And there it's also static (but for a reason). 
Here it's not necessary. But still possible - and if you don't mind - I would prefer to keep it - only because of "convention reasons".

app/Filter/Forms/Types/AuthoritySearchRequestType.php
igor-kamil: I decided to get ispired by the naming/convention in [symfony forms](https://symfony.com/doc/current/reference/forms/types/form.html#translation-domain)
Initially I wanted to use it directly - since there is `choice_translation_domain` option available (but it's not possible to use in in our case - since we mix up the choice values with counts)
We already use this option e.g. in https://github.com/SlovakNationalGallery/webumenia.sk/blob/feature/WEBUMENIA-1584-add-gender-filter-other-variant/app/Filter/Forms/Types/AuthoritySearchRequestType.php#L92

But now I see - there it's option. Here it's variable name. Do you think it would be better to use camelCase?

app/Elasticsearch/Repositories/ItemRepository.php
```diff
@@ -168,7 +168,7 @@ public function getSimilarByColor(int $size, Item $item, $locale = null): Search
         return $this->createSearchResult($response);
     }
 
-    protected function createBucketCollection(array $response, string $attribute): Collection
+    protected function createBucketCollection(array $response, string $attribute, ?string $translation_domain = null): Collection
```
eronisko, line 171: Does the `$translation_domain` get used here at all?

app/Filter/Forms/Types/AuthoritySearchRequestType.php
eronisko: On a second look, I'm wondering whether the translation domain should be mentioned in repositories at all. Something like `authorityRepository->listValues` should probably not be concerned with the presentation at all, IMHO. Same for `createBucketCollection()`. I think we're mixing in presentation concerns too soon.

Another thing is that `TranslatableRepository->listValues` now has both `$locale` and `$translation_domain`. That makes things a bot confusing, doesn't it? Now I see that this was probably originally designed to store localized strings in the index as proposed in #596.

My proposal would be to limit the translation domain logic only to this function `$getChoiceOptions`. We'd probably have to refactor the `listValues` to return something like 
```php
[
  [
    'value' => 'ABC',
    'count' => 33,
  ],
  [
    'value' => 'XYZ',
    'count' => 14,
  ],
]
```
and to turn that into actual [choices](https://symfony.com/doc/current/reference/forms/types/choice.html#choices) here, in `$getChoiceOptions`. 

That would reduce the presentation logic "creep" from Repositories and it would give us more flexibility to translate closer to the front-end.

eronisko: Also, if you'll want to revert to #596, that would be fine by me too. It would be the cheaper/faster option.

eronisko (approved):

PR #606, opened 2022-01-14
GMUHK authority matching

rastislav-chynoransky:
# Description

Fixes https://jira.sng.sk/browse/WEBUMENIA-1651

## Type of change

# Checklist:

### DIFF
--- a/app/Authority.php
+++ b/app/Authority.php
@@ -122,6 +122,14 @@ class Authority extends Model implements IndexableModel, TranslatableContract
         return $this->links()->where('type', 'source');
     }
 
+    public function scopeSwappedName($query, $name)
+    {
+        $query->where(
+            DB::raw('concat(substring_index(name, ", ", -1), " ", substring_index(name, ", ", 1))'),
+            $name
+        );
+    }
+
     public function getCollectionsCountAttribute()
     {
         if (!Cache::has('authority_collections_count')) {
--- a/app/Authority.php
+++ b/app/Authority.php
@@ -122,11 +122,13 @@ class Authority extends Model implements IndexableModel, TranslatableContract
         return $this->links()->where('type', 'source');
     }
 
-    public function scopeFirstNameLastName($query, $name)
+    public function scopeWhereFirstNameLastName($query, $name)
     {
-        $query->where(
-            DB::raw('concat(substring_index(name, ", ", -1), " ", substring_index(name, ", ", 1))'),
-            $name
+        $query->whereRaw(
+            'case when instr(name, ", ") then ' .
+            'concat(substring_index(name, ", ", -1), " ", substring_index(name, ", ", 1)) = ? ' .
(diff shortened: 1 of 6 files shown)


app/Authority.php
```diff
@@ -122,6 +122,14 @@ public function sourceLinks()
         return $this->links()->where('type', 'source');
     }
 
+    public function scopeSwappedName($query, $name)
```
eronisko, line 125: I would suggest adding `where` to the scope name so that we know that it's a filter

```suggestion
    public function scopeWhereSwappedName($query, $name)
```

I'm also not sure if _swapped_ is the right name if you don't know what "unswapped" looks like.

```suggestion
    public function scopeWhereFirstNameLastName($query, $firstNameLastName)
```

eronisko (approved):

app/Authority.php
igor-kamil: imho "scope" already suggest "where". also using `->whereSomehting()` might conflict with native magical `->whereAttribute('')` (I can't find it in documentation -> I mean this https://laracasts.com/discuss/channels/eloquent/whereattribute-function?page=1&replyId=402083 )

igor-kamil (approved):
for me OK

app/Authority.php
eronisko: > imho "scope" already suggest "where"

In the definition in does, but in the usage not necessarily. Compare e.g. 

```php
Authority::query()->swappedName($fullname)`
```
vs
```php
Authority::query()->whereSwappedName($fullname)`
```

scopes without a preposition suggest a trait to me, e.g. "Authorities that have a swapped name" rather than "Authorities where swapped name is XYZ".

I think overriding the `whereAttribute` is completely fine here, since we don't have such attribute anyway.

rastislav-chynoransky:
I've incorporated your suggestions and added one more use case. Please check.

app/AuthorityName.php
```diff
@@ -20,4 +18,14 @@ public function authority()
...
     }
+
+    public function scopeWhereFirstNameLastName($query, $name)
+    {
+        $query->whereRaw(
+            'case when instr(name, ", ") then ' .
+            'concat(substring_index(name, ", ", -1), " ", substring_index(name, ", ", 1)) = ? ' .
+            'else name = ? end',
+            [$name, $name]
+        );
```
eronisko, line 29: Would this also work?


```suggestion
        $query
            ->where('name', $name)
            ->orWhereRaw(
                'concat(substring_index(name, ", ", -1), " ", substring_index(name, ", ", 1)) = ? ',
                $name
            );
```

eronisko (approved):

app/AuthorityName.php
rastislav-chynoransky: Sure, but I found it a bit ignorant, but I know that the usage of `case` is less readable...

PR #607, opened 2022-01-20
RFC: Tailwind experiment

eronisko:
@igor-kamil @rastislav-chynoransky 
While working on the new homepage, it occurred to me that we could probably already be using Tailwind on new pages.

I've put together a config that I think will work alongside the current Bootstrap set-up as long as we:

- use a `tw-` prefix for Tailwind classes (this works even with IDE completion): https://tailwindcss.com/docs/configuration#prefix
- keep new pages/component within `.tailwind-rules` classes (https://tailwindcss.com/docs/configuration#important). This is because we have to disable certain defaults e.g. for headings (see app-tailwind.css).

So this proposed setup not ideal, but it is something that would allow us to use Tailwind today without committing to a full-on migration.

I've deployed https://test.webumenia.sk/new-home so that you can see this in action. Right now this adds ~3KB to our bundle.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ public/sitemap.xml
 /public/css/admin.css
 /public/css/mail
 /public/css/style.css
+/public/css/app-tailwind.css
 /public/fonts
 /public/hot
 /public/js/*LICENSE.txt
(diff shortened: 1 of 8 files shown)


igor-kamil (commented):
my initial feelings were bit skeptical (asking, whether this is not messing up things even more). 
on the second sight I like how it allows us to easily use tailwind - without the necessity to refactor whole frontend code. 
also later it still allows us to make full switch from BS to TW in the future. 
this is good candidate for current situation. but let's consider also other possibilities.

igor-kamil (approved):
OK for me

rastislav-chynoransky (approved):

eronisko:
Thank you both

PR #611, opened 2022-01-28
Basic Tailwind set-up + WIP new homepage 

eronisko:
# Description
image

This is WIP for the new 'latest content' section.

The section is not yet done but I wanted to merge this in, as it contains a number of other changes and I didn't want the branch to get stale. The changes are:

* Basic Tailwind set-up working within the `.tailwind-rules` as seen in #607 
* Basic color and font size config similar to what we have today
* Tabs Vue component, inspired by [Headless UI Tabs](https://headlessui.dev/vue/tabs) (unfortunately, we cannot use Headless as it depends on Vue3
* a simple Flickity wrapper to component to configure flickity from within HTML

## Type of change

# Checklist:


resources/css/app-tailwind.css
```diff
@@ -0,0 +1,12 @@
+@import "tailwindcss/base";
+@import "./tailwind-preflight-scoped.css";
+
+@tailwind components;
+@tailwind utilities;
+
+@layer base {
+    html {
+        /* Set rem size to 14px */
+        font-size: 14px !important;
```
igor-kamil, line 10: is this really necessary?

resources/css/tailwind-preflight-scoped.css
```diff
@@ -0,0 +1 @@
...es h1,.tailwind-rules h2,.tailwind-rules h3,.tailwind-rules h4,.tailwind-rules h5,.tailwind-rules h6,.tailwind-rules hr,.tailwind-rules p,.tailwind-rules pre{margin:0}.tailwind-rules fieldset{margin:0;padding:0}.tailwind-rules legend{padding:0}.tailwind-rules menu,.tailwind-rules ol,.tailwind-rules ul{list-style:none;margin:0;padding:0}.tailwind-rules textarea{resize:vertical}.tailwind-rules input::-moz-placeholder,.tailwind-rules textarea::-moz-placeholder{color:theme("colors.gray.400",#9ca3af);opacity:1}.tailwind-rules input:-ms-input-placeholder,.tailwind-rules textarea:-ms-input-placeholder{color:theme("colors.gray.400",#9ca3af);opacity:1}.tailwind-rules input::placeholder,.tailwind-rules textarea::placeholder{color:theme("colors.gray.400",#9ca3af);opacity:1}.tailwind-rules [role=button],.tailwind-rules button{cursor:pointer}.tailwind-rules :disabled{cursor:default}.tailwind-rules audio,.tailwind-rules canvas,.tailwind-rules embed,.tailwind-rules iframe,.tailwind-rules img,.tailwind-rules object,.tailwind-rules svg,.tailwind-rules video{display:block;vertical-align:middle}.tailwind-rules img,.tailwind-rules video{height:auto;max-width:100%}.tailwind-rules [hidden]{display:none}
```
igor-kamil, line 1: this is generated file, isn't it? do we need to keep it within version control?

igor-kamil (approved):
overall I like it. I think this is nice step to use tailwind - without really breaking anything.
please check - within flickity carousel, the font sizes for icons became slightly bigger:
(current vs new)
Screenshot 2022-02-02 at 23 35 35
Screenshot 2022-02-02 at 23 35 42

on my local machine `npm run watch` leads to infinite loop 
Screenshot 2022-02-02 at 20 49 25
(tested with `npm v8.1.2` and `npm v6.14.15`)

resources/css/app-tailwind.css
eronisko: I think so, but it's probably not explained very well...

The Tailwind's spacing scale works in rems, and it was set up with [`1rem = 16px` in mind](https://tailwindcss.com/docs/customizing-spacing#default-spacing-scale). Both the rem size and the spacing scale are obviously adjustable, but Bootstrap [sets 1rem to 10px](https://github.com/SlovakNationalGallery/webumenia.sk/blob/958aeb32c741977d71d0b53b2d272e27a9182158/resources/less/bootstrap/scaffolding.less#L23), which just made Tailwind's default spacing scale very small. This meant we'd have to customize the scale pretty heavily, which is probably not trivial to get right.

So I opted for our current body font size, 14px. That way, we get to be propotional to the font sizes and line heights, and have a scale that is easily translatable between other Tailwind projects.

resources/css/tailwind-preflight-scoped.css
eronisko: You're right, it is a generated file, here's why I chose to check it in:

1. At first I thought we could just cherry-pick rules from the Preflight stylesheet, but I soon realized that it's really hard to tell which ones are necessary. So I decided it'll be best to get them all.
1. I wanted the rules to be scoped by `.tailwind-rules` (so that they don't interfere with out current stylesheet),  but I didn't want to re-write Preflight by hand. So I utilized Less to do the scoping.
1. At the same time, I wanted to keep the Tailwind pipeline as standard as possible, and since they [advise against using other preprocessors](https://tailwindcss.com/docs/using-with-preprocessors#using-sass-less-or-stylus), I wanted to keep Tailwind stuff PostCSS-only.
1. I found it tricky to "pipe" less output to PostCSS via Mix (Mix is really geared toward outputing everything into /public), so I came up with this solution: keep the Less file as a guide, use the generated CSS in PostCSS/Mix.

I believe this set-up will be OK in the end. We'll only ever need to touch it if we update Tailwind _and_ it's still a temporary workaround until we remove Bootstrap.

eronisko:
@igor-kamil @rastislav-chynoransky I've had to make a few more adjustments to existing CSS in order to accommodate for the new rem size https://github.com/SlovakNationalGallery/webumenia.sk/pull/611/commits/b9b24f79f919df42cf3abd50f3105f259d420fcb as per @igor-kamil's observation in the review.

These are all minor & simple adjustments do css classes so I hope it's OK if I go ahead and merge this in. If you take a look and find anything you don't like I'll be happy to revert & fix this commit later.

eronisko:
@igor-kamil as per your issue with `npm run watch`, I have not been able to reproduce. @rastislav-chynoransky, does this happen to you too?

PR #620, opened 2022-02-09
[item] use authority matcher in indexing

rastislav-chynoransky:
# Description

Fixes https://jira.sng.sk/browse/WEBUMENIA-1777

## Type of change

# Checklist:

### DIFF
--- a/app/Item.php
+++ b/app/Item.php
@@ -351,13 +351,6 @@ class Item extends Model implements IndexableModel, TranslatableContract
             ->keys();
     }
 
-    public function getMergedAuthorityNamesAndAuthors()
-    {
-        return $this->authorities
-            ->pluck('name')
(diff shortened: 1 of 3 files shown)


app/Item.php
```diff
@@ -653,7 +660,7 @@ public function getIndexedData($locale)
         return [
             'id' => $this->id,
             'identifier' => $this->identifier,
-            'author' => $this->makeArray($this->author),
+            'author' => $this->makeArray($this->getMergedAuthorityNamesAndAuthors()->toArray()),
```
eronisko, line 663: Is it worth making this into a public function if it's only being used here? Do you see it used elsewhere in the future?

If not, I would make it private/inline and test it against `getIndexedData()`

eronisko (changes requested):

eronisko (approved):

PR #622, opened 2022-02-10
Disable success notifications in Mix

eronisko:
Because working with Tailwind's JIT compiler, there are a lot of them

https://laravel-mix.com/docs/6.0/os-notifications

### DIFF
--- a/webpack.mix.js
+++ b/webpack.mix.js
@@ -46,7 +46,8 @@ mix
     .postCss('resources/css/app-tailwind.css', 'public/css', [
         require('tailwindcss')
     ])
-
+    
+    .disableSuccessNotifications()
     .options({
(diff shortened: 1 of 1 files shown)


igor-kamil (commented):
I am not sure. I find even "success" notification useful (often I wait to refresh the browser until I see it got compiled. for scripts it's instant - but for styles it takes some time)
I don't want to block it. I really need only "failing" notification. So I let up to you and @rastislav-chynoransky

eronisko:
@igor-kamil FWIW, you'll still get the "first" success notification. So one notification after starting the script and then first success notification after a failed build.

igor-kamil (approved):
ok. now I see your point.

PR #637, opened 2022-02-24
Latest content design tweaks

eronisko:
* larger container size
* better card sizing
* re-ordered card elements

image

image


igor-kamil (approved):
LGTM 
(i18n would be probably next step, right?)

PR #638, opened 2022-02-24
Fix & streamline font loading

eronisko:
Currently, there are no assets served for Source Serif Pro:
image

Changes:
* SourceSerifPro is served over network
* changed from Less to css: original `less/fonts.less` is split into `css/fonts.css` and `less/icons.less`
* An unused `less/fonts.css` removed

## Type of change

# Checklist:


resources/css/fonts.css
```diff
@@ -0,0 +1,273 @@
+@font-face {
+  font-family: "Source Serif Pro";
+  font-weight: 200;
+  font-display: swap;
+  font-style: normal;
+  font-stretch: normal;
+  src: url("../fonts/SourceSerifPro/WOFF2/TTF/SourceSerifPro-ExtraLight.ttf.woff2")
```
igor-kamil, line 7: I would consider using `@import` from google fonts -> https://fonts.google.com/specimen/Source+Serif+Pro
also please load only used styles... if possible

resources/less/fonts.css
```diff
@@ -1,15 +0,0 @@
-/* Generated by Font Squirrel (http://www.fontsquirrel.com) on April 28, 2014 */
-@font-face {
-  font-family: 'futura_t';
```
igor-kamil, line 3: `futura_t` isn't used anywhere? how it even ever appeared here? :D

resources/less/style.less
```diff
@@ -5,7 +5,8 @@
 
 //@import "variables.less";
 @import "responsive-align.less";
-@import "fonts.less";
+@import "../css/fonts.css";
+@import "icons.less";
```
igor-kamil, line 9: I appreciate separation fonts and icons
I don't know the reason to move from `fonts.less` to `fonts.css`, till we use `less` everywhere else anyway. But I don't mind. I am just bit curious 😛

igor-kamil (commented):
nice catch 🕶️ 
thank you for the fix
please see my comments

igor-kamil (approved):
I tested it locally. and it's good as it's. 
you were right - it loads only styles being used. and even I have already copied `/fonts/SourceSerifPro`
not needed to do anything more here

resources/less/style.less
eronisko: I've moved it to fonts.css so that I can use it in Tailwind: 

https://tailwindcss.com/docs/using-with-preprocessors#using-sass-less-or-stylus
> For the best development experience, we highly recommended that you [use PostCSS exclusively](https://tailwindcss.com/docs/using-with-preprocessors#using-post-css-as-your-preprocessor), and that you don’t use preprocessors like Sass or Less in your Tailwind projects.

resources/less/fonts.css
eronisko: 🤷‍♂️

PR #645, opened 2022-03-04
"Featured artwork" front-end 

eronisko:
image
image

## Type of change

# Checklist:

### DIFF
--- a/resources/css/app-tailwind.css
+++ b/resources/css/app-tailwind.css
@@ -6,18 +6,23 @@
 @tailwind utilities;
 
 @layer base {
+    /* Resets that are not covered by Preflight */
+    /* TODO remove after retiring Bootstrap */
     html {
-        /* TODO remove after retiring Bootstrap */
         font-size: 16px !important;
+    }
+
+    .tailwind-rules a:hover, .tailwind-rules a:focus {
+        @apply tw-text-inherit tw-decoration-inherit;
     }
 
     /* Theme defaults */
     /* Once Bootstrap is removed, the applied classes here can be moved to <body> */
     .tailwind-rules {
-        @apply tw-text-base tw-text-gray-800;
+        @apply tw-text-base tw-text-gray-800 tw-font-sans;
     }
     .tailwind-rules.admin {
-        @apply tw-text-sm;
+        @apply tw-text-sm tw-font-[Arial];
     }
 }
 
--- a/resources/css/app-tailwind.css
+++ b/resources/css/app-tailwind.css
@@ -12,7 +12,8 @@
         font-size: 16px !important;
     }
 
-    .tailwind-rules a:hover, .tailwind-rules a:focus {
+    .tailwind-rules a:hover,
+    .tailwind-rules a:focus {
         @apply tw-text-inherit tw-decoration-inherit;
     }
 
@@ -22,7 +23,8 @@
         @apply tw-text-base tw-text-gray-800 tw-font-sans;
     }
     .tailwind-rules.admin {
-        @apply tw-text-sm tw-font-[Arial];
+        @apply tw-text-sm;
+        font-family: 'Helvetica Neue', Helvetica, Arial, sans-serif;
     }
 }
 
--- a/resources/views/components/item_image.blade.php
+++ b/resources/views/components/item_image.blade.php
@@ -1,0 +1,8 @@
+@props(['id'])
+
+<img {{ $attributes->merge([
+  'src' => route('dielo.nahlad', ['id' => $id, 'width' => 600]),
+  'srcset' => collect([220, 300, 600, 800])
+    ->map(fn ($width) => route('dielo.nahlad', ['id' => $id, 'width' => $width]) . " ${width}w")
+    ->join(', ')
+]) }} />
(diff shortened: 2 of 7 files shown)


resources/views/components/item_image.blade.php
```diff
@@ -0,0 +1,8 @@
+@props(['id'])
+
+<img {{ $attributes->merge([
+  'src' => route('dielo.nahlad', ['id' => $id, 'width' => 600]),
+  'srcset' => collect([220, 300, 600, 800])
+    ->map(fn ($width) => route('dielo.nahlad', ['id' => $id, 'width' => $width]) . " ${width}w")
+    ->join(', ')
```
igor-kamil, line 7: Collections magic again! ✨ 
With your permission - I am going to reuse this in bookmarks 😁

resources/views/home/index.blade.php
```diff
@@ -34,16 +34,59 @@
...
+            <div class="tw-text-white tw-bg-gray-800 tw-py-8 lg:tw-py-16">
+                <div class="tw-container tw-mx-auto tw-px-6 tw-max-w-screen-xl tw-grid lg:tw-grid-cols-2 lg:tw-gap-x-14">
+                    <div class="lg:tw-order-1 tw-text-center">
+                        <a href="{{ route('dielo', ['id' => $featuredArtwork->item->id]) }}" class="tw-inline-block">
+                            <x-item_image :id="$featuredArtwork->item->id" class="tw-max-h-80 lg:tw-max-h-[32rem]" />
+                        </a>
+                    </div>
+                    <h2 class="lg:tw-text-2xl tw-font-semibold tw-mt-6 tw-mb-2 lg:tw-mt-0 lg:tw-mb-6 lg:tw-col-span-2">
+                        Dielo dňa
+                    </h2>
```
igor-kamil, line 60: Is there any plan (in the future) to refactor pieces of code like this into reusable components? (every time when I see _way-too-many-classes-used-within-one-header_ I feel like "I will name you: "_______")
I guess similar code (with +/- same classes) is also used for featured collection/article.

resources/css/app-tailwind.css
```diff
@@ -6,18 +6,23 @@
...
+
     /* Theme defaults */
     /* Once Bootstrap is removed, the applied classes here can be moved to <body> */
     .tailwind-rules {
-        @apply tw-text-base tw-text-gray-800;
+        @apply tw-text-base tw-text-gray-800 tw-font-sans;
     }
     .tailwind-rules.admin {
-        @apply tw-text-sm;
+        @apply tw-text-sm tw-font-[Arial];
```
igor-kamil, line 25: why? and why arial? 
to be honest - I would love to unify font families in admin with the ones used on frontend (btw - check login screen - already there - smooth and nice)
I know it has zero priority. but same has enforcing arial in my point of view.
(and I see admin in Helvetica - what is the first option )

igor-kamil (approved):
looks nice. I really like the new `item_image` component. 
no issues from my point of view. just few (low priority) questions.

resources/views/home/index.blade.php
eronisko: There are some reusable pieces here, but probably not that many. For example, I'm not sure if making this heading into a component would be a good choice: we don't necessarily want to couple this particular style to the `<h2>` tag, and extracting the style into a custom class is something the Tailwind docs recommend against:

See for example https://tailwindcss.com/docs/reusing-styles#avoiding-premature-abstraction:

>Whatever you do, don’t use @apply just to make things look “cleaner”. Yes, HTML templates littered with Tailwind classes are kind of ugly. Making changes in a project that has tons of custom CSS is worse.

Whenever I feel like extracting something, I read through that page 😄 Note though, I'm still working in a single file, so multi-cursor/looping is fine for now. Once we start re-using those styles in other templates, it'll become more cumbersome and the right candidates for extraction will crop up. 

I also quite like the approach here, extracting the "function" rather than the style into components: https://youtu.be/nqNIy8HkEQ8?t=880

I know this is different to the way this was done in Komarch, but so far this has worked well for me and copy-pasting is actually fine so far.

Two more things might make this better:
1. Having the Figma design match our Tailwind config really helps. That way I don't have to think about "What component is this? Heading large? Subtitle?", I can just look at the font size/weight/etc and type it out, it's pretty easy with Tailwind intellisense. And we're 95% there already.
1. Sorting the CSS classes so that it's easier to spot patterns in code. This [Prettier plugin for Blade](https://github.com/shufo/prettier-plugin-blade) got Tailwind CSS sorting 9 days ago, so I'll try to plug it in (in a separate PR)

resources/css/app-tailwind.css
eronisko: What I tried to do here was to match the current style. On Windows it's Arial and I didn't bother looking at the CSS, so this is a mistake. It should be `"Helvetica Neue", Helvetica, Arial, sans-serif`. Will fix.

resources/views/home/index.blade.php
igor-kamil: 👌 I agree. thank you for such a comprehensive answer 😉

PR #646, opened 2022-03-10
"Featured author" front-end component

eronisko:
image
image

## Type of change

# Checklist:

### DIFF
--- a/app/Http/Controllers/HomeController.php
+++ b/app/Http/Controllers/HomeController.php
@@ -3,6 +3,7 @@
 namespace App\Http\Controllers;
 
 use App\Article;
+use App\Authority;
 use App\Collection;
 use App\FeaturedArtwork;
 use App\FeaturedPiece;
@@ -21,6 +22,13 @@ class HomeController extends Controller
         $featuredArtwork = FeaturedArtwork::query()
             ->published()
             ->orderBy('updated_at', 'desc')
+            ->first();
+
+        $featuredAuthor = Authority::query()
+            ->where('has_image', true)
+            ->where('type', 'person')
+            ->has('items', '>', 3)
+            ->inRandomOrder()
             ->first();
 
         $articles = Article::query()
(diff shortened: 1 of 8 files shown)


app/Http/Controllers/HomeController.php
```diff
@@ -23,6 +24,13 @@ public function index()
             ->orderBy('updated_at', 'desc')
             ->first();
 
+        $featuredAuthor = Authority::query()
+            ->where('has_image', true)
+            ->where('type', 'person')
+            ->has('items', '>', 3)
```
igor-kamil, line 30: "3"? set as constant. or what about `scopeFeatured()` in authority model?

app/Http/Controllers/HomeController.php
```diff
@@ -23,6 +24,13 @@ public function index()
...
+            ->inRandomOrder()
```
igor-kamil, line 31: this might be very slow. isn't it? 
combination of `where` and `order by rand()` is expensive (related discussion: https://stackoverflow.com/questions/1594483/optimizing-my-mysql-statement-rand-too-slow )
that's why we used elastic search in https://github.com/SlovakNationalGallery/webumenia.sk/blob/develop/app/Exceptions/Handler.php#L63-L66

resources/views/home/index.blade.php
```diff
@@ -122,8 +122,7 @@ class="tw-hidden sm:tw-inline-block tw-text-sm tw-border-gray-300 tw-border tw-p
                 </div>
                 <div class="tw-mt-6 md:tw-mt-8">
                     <tab-panel v-slot="{ active }" class="tw-relative">
-                        <flickity :resize-once="active"
-                            :options="{ cellAlign: 'left', contain: true, pageDots: false, prevNextButtons: false, freeScroll: true }">
+                        <x-home.carousel v-bind:resize-once="active">
```
igor-kamil, line 125: 👏

igor-kamil (commented):
everything looks fine. I like the new `carousel` component. 
I have just bit concerns regarding performance of querying random featured authors
(we were dealing with this issue in the past with "random featured artwork")
my suggestion is to use elastic for such a case

app/Http/Controllers/HomeController.php
eronisko: I just made this up for now, this is to be changed for sure.

app/Http/Controllers/HomeController.php
eronisko: https://jira.sng.sk/browse/WEBUMENIA-1799

eronisko: Deployed to https://test.webumenia.sk/new-home, perf seems fine to me but I've created a performance check-up task to be sure :)

igor-kamil (approved):

PR #652, opened 2022-03-12
[item] resolve api route name conflicts

rastislav-chynoransky:
Currently, there is a conflict with our proxy.
How would you resolve?
I'm open to your suggestions.


eronisko:
I'm happy with the proposed changes, it's a good solution. I'd just consider deprecating/moving the current API in favour of this one.

eronisko (approved):

PR #653, opened 2022-03-15
add word-break:keep-all property globally

rastislav-chynoransky:
# Description

Fixes https://jira.sng.sk/browse/PNP-71

## Type of change

# Checklist:

### DIFF
--- a/resources/css/app-tailwind.css
+++ b/resources/css/app-tailwind.css
@@ -21,6 +21,7 @@
     /* Once Bootstrap is removed, the applied classes here can be moved to <body> */
     .tailwind-rules {
         @apply tw-text-base tw-text-gray-800 tw-font-sans;
+        word-break: keep-all;
     }
     .tailwind-rules.admin {
         @apply tw-text-sm;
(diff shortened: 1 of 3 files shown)


resources/css/app-tailwind.css
```diff
@@ -21,6 +21,7 @@
     /* Once Bootstrap is removed, the applied classes here can be moved to <body> */
     .tailwind-rules {
         @apply tw-text-base tw-text-gray-800 tw-font-sans;
+        word-break: keep-all;
```
eronisko, line 24: Please make a class for this on the `utilities` layer and then `@apply` it so that it can be used on `<body>` once we retire Bootstrap (Although I can see I didn't make a very good example here with the `.tailwind-rules.admin` class 😅)

Still, I'm kinda hesitant with making this global. I'd probably just make a utility class (even outside of Tailwind if it's easier) and apply it just where it matters, e.g. image thumbnails, descriptions, title. What does @igor-kamil think?

eronisko (changes requested):

resources/css/app-tailwind.css
eronisko: Btw, these artworks are 😍

rastislav-chynoransky: Since this setting is the same as `normal` (default) and only affects CJK languages I'd prefer it to have set globally as there may be many places to apply it to (even in the future development).

eronisko (approved):

PR #654, opened 2022-03-17
Added "counts blurb" to new homepage

eronisko:
image

## Type of change

# Checklist:

### DIFF
--- a/app/Http/Controllers/HomeController.php
+++ b/app/Http/Controllers/HomeController.php
@@ -73,7 +73,7 @@ class HomeController extends Controller
 
     private function getCountsBlurbData()
     {
-        [$itemsCount, $choices] = Cache::remember('home.counts', now()->addDays(3), function () {
+        $choices = Cache::remember('home.counts', now()->addDays(3), function () {
             $galleriesCount = 15;
             $authoritiesCount = Authority::count();
             $itemsCount = Item::count();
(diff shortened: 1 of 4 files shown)


rastislav-chynoransky (approved):

app/Http/Controllers/HomeController.php
```diff
@@ -61,7 +66,56 @@ public function index()
...
                 'collectionsRemainingCount',
+                'countsBlurb',
             ])
         );
     }
+
+    private function getCountsBlurbData()
+    {
+        [$itemsCount, $choices] = Cache::remember('home.counts', now()->addDays(3), function () {
+            $galleriesCount = 15;
```
igor-kamil, line 77: btw. it's count of https://github.com/SlovakNationalGallery/webumenia.sk/blob/develop/routes/web.php#L344-L430
would be nice to handle this somehow in the future...

app/Http/Controllers/HomeController.php
```diff
@@ -61,7 +66,56 @@ public function index()
...
+            $authoritiesCount = Authority::count();
+            $itemsCount = Item::count();
+            $highResItemsCount = Item::withExists('images')->count();
```
igor-kamil, line 80: any reason why not to "recycle" good old `$itemRepository->count((new ItemFilter)->setHasIip(true))`?

app/Http/Controllers/HomeController.php
```diff
@@ -61,7 +66,56 @@ public function index()
...
+            $freeItemsCount = $itemRepository->count((new ItemFilter())->setIsFree(true));
+
+            return [
+                $itemsCount,
+                [
+                    [
+                        trans('intro.from_galleries_start'),
+                        route('frontend.info'),
+                        $galleriesCount,
+                        trans('intro.from_galleries_end'),
```
igor-kamil, line 92: I understand it comes up from the legacy code. But now I feel I am missing keys in this array. 
Also I would prefer to see `trans` used in view, not in controller. 
Something like

```suggestion
                        'claim_start' => 'intro.from_galleries_start',
                        'url' => route('frontend.info'),
                        'count' => $galleriesCount,
                        'claim_end' => 'intro.from_galleries_end',
```

resources/views/home/index.blade.php
```diff
@@ -49,6 +49,24 @@ class="tw-relative tw-mt-3 tw-inline-block tw-border tw-border-gray-300 tw-px-4
...
+        {{-- Counts blurb --}}
+        <div class="tw-bg-gray-200">
+            <div
+                class="tw-container tw-mx-auto tw-grid tw-max-w-screen-xl tw-px-6 tw-py-5 tw-text-gray-500 lg:tw-py-10">
+                <p class="tw-text-center lg:tw-text-2xl">
+                    {{ utrans('intro.definition_start') }}
+                    <a href="{{ route('frontend.catalog.index') }}"
+                        class="tw-font-bold tw-text-gray-800 hover:tw-underline">{!! formatNum($countsBlurb->itemsCount) !!}</a>
+                    {{ trans('intro.definition_end') }}<br>
+                    {{ $countsBlurb->parts[0] }}
```
igor-kamil, line 62: from my point of view - this is not very intuitive. to use "part + array index". at least I would prefer to make it "associated array" - to use string instead of number. something like
```suggestion
                    {{ trans($countsBlurb->parts['claim_start']) }}
```

igor-kamil (approved):
I am ok with this as it is. Since it works nicely and more-or-less reuse the existing code.  
Still I feel like this is also an opportunity to make it in little bit more elegant way.
Anyway - we can leave it for the next time...

app/Http/Controllers/HomeController.php
eronisko: Good suggestion, I'll do this in a separate PR

app/Http/Controllers/HomeController.php
eronisko: I just prefer not to rely on ES for things like these. Performance-wise this seems fine even if when uncached (I did check).

app/Http/Controllers/HomeController.php
eronisko: Yeah, I agree. I just wanted to re-use the old solution but it was kinda confusing. Changed now.

PR #655, opened 2022-03-17
Caching for new front-page

eronisko:
## Type of change

# Checklist:

### DIFF
--- a/app/Collection.php
+++ b/app/Collection.php
@@ -9,6 +9,7 @@ use Carbon\Carbon;
 use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Database\Query\JoinClause;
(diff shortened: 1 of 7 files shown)


rastislav-chynoransky (approved):

app/Collection.php
```diff
@@ -17,6 +18,11 @@ class Collection extends Model implements TranslatableContract
 
     use HasHeaderImage;
 
+    protected static function booted()
+    {
+        static::saved(fn() => Cache::forget('home.collections'));
+    }
+
```
igor-kamil, line 25: use Trait for this?

igor-kamil (approved):
🚀 👌 
(the idea with trait is just optional...)

app/Collection.php
eronisko: I'd guess turning it into a trait would require more code than having it out in the open like this.

PR #656, opened 2022-03-17
Linting

eronisko:
Files re-formatted with prettier. No other changes.

### DIFF
--- a/app/Article.php
+++ b/app/Article.php
@@ -1,7 +1,5 @@
 <?php
 
-
-
 namespace App;
 
 use App\Concerns\HasHeaderImage;
@@ -33,15 +31,9 @@ class Article extends Model implements TranslatableContract
 
     public $translatedAttributes = ['title', 'summary', 'content'];
 
-    protected $fillable = [
-        'published_date',
-    ];
+    protected $fillable = ['published_date'];
 
-    protected $dates = [
-        'created_at',
-        'updated_at',
-        'published_date'
-    ];
+    protected $dates = ['created_at', 'updated_at', 'published_date'];
 
     protected $casts = [
         'edu_media_types' => 'array',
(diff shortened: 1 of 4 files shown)


rastislav-chynoransky:
Is this applied to all files? Don't we want to make it automatic for every commit?

eronisko:
@rastislav-chynoransky Prettier is typically used to format your code when you save a file so that you don't have to worry about formatting while working. For me, not having to worry about spaces, newlines etc while writing code is the main benefit; having all files neat and tidy is secondary.

I've considered doing a 'format all' commit but decided against it in the end (it would obscure some of our commit history and the benefit is... 🤷‍♂️). But I'd be happy to do it if you guys are up for it.

In this particular PR I wanted to separate the re-formatting from the changes in other PRs, just so that they're easier to digest.

BTW, I'm sorry I never explained the inclusion of Prettier to you. I don't know how familiar are you with it and how much do you use auto-formatting in your IDE. For me, "format on save" has been a great time saver, so I highly recommend that setup :) https://prettier.io/

rastislav-chynoransky:
Thanks for the explanation. I haven't noticed it being pre-configured for the project. I'll be happy to use it.
However when I want to enable it in vscode for modifications only it fails with error. Have you had the same issue?

rastislav-chynoransky (approved):

eronisko:
@rastislav-chynoransky 
> However when I want to enable it in vscode for modifications only it fails with error. Have you had the same issue?

I haven't had much success with formatting "modifications only" in general. Maybe the Output view will give you some hints:
image

If the re-format is too big a change, there's always the ["save without formatting"](https://glebbahmutov.com/blog/configure-prettier-in-vscode/#saving-without-formatting) option.

igor-kamil (approved):

PR #658, opened 2022-03-18
Unset global a:hover a:focus styles

eronisko:
This change removes all global (unscoped) `a:hover` and `a:focus` styles that have been introduced by either Bootstrap or our own CSS. I could not find any other way to re-set these styles in a scoped CSS so that we could use Tailwind reliably (within `.tailwind-rules`).

This is potentially a breaking change, but I've tested links on all pages I could think of and did not find this change to have any effect -- `<a>` tags in the wild are usually overriden within other scopes.

## Type of change

# Checklist:

### DIFF
--- a/resources/css/app-tailwind.css
+++ b/resources/css/app-tailwind.css
@@ -16,12 +16,6 @@
         transition: unset;
     }
 
-    .tailwind-rules a:hover,
-    .tailwind-rules a:focus {
-        @apply tw-decoration-inherit;
-        color: initial;
-    }
-
     /* Theme defaults */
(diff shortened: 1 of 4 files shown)


resources/css/app-tailwind.css
```diff
@@ -13,13 +13,7 @@
...
 
     .tailwind-rules a {
-        transition: unset;
-    }
-
-    .tailwind-rules a:hover,
-    .tailwind-rules a:focus {
-        @apply tw-decoration-inherit;
-        color: initial;
+        transition: inherit;
```
igor-kamil, line 16: much better!

igor-kamil (approved):
🤞

PR #661, opened 2022-03-18
Import style period, current location

rastislav-chynoransky:
# Description

Fixes https://jira.sng.sk/browse/WUCZ-3

## Type of change

# Checklist:

### DIFF
--- a/resources/lang/sk.json
+++ b/resources/lang/sk.json
@@ -1,0 +1,5 @@
+{
+    "importer.cs.style_period.České umění 19. st.": "České umenie 19. st.",
+    "importer.cs.style_period.České umění 20. st.": "České umenie 20. st.",
+    "importer.cs.style_period.České umění 21. st.": "České umenie 21. st."
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@ All notable changes to this project will be documented in this file[^1].
 
 ## [Unreleased]
 ### Added
-- item style period
+- item style period, current location
 
 ## [2.47.0] - 2022-03-11
 ### Added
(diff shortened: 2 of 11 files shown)


database/migrations/2022_03_18_163014_add_style_period_and_location_to_item_translations_table.php
```diff
@@ -0,0 +1,33 @@
...
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('item_translations', function (Blueprint $table) {
+            $table->string('style_period')->nullable();
+            $table->string('location')->nullable();
```
igor-kamil, line 18: I was wondering, what's the difference between `location` and `place`... but now I see -  "_current location_"

igor-kamil (approved):
🆗

database/migrations/2022_03_18_163014_add_style_period_and_location_to_item_translations_table.php
rastislav-chynoransky: Good point... Should i rename those to `current_location` and `origin_location`?

igor-kamil: `current_location` sounds better 👍

resources/lang/sk.json
```diff
@@ -0,0 +1,5 @@
+{
+    "importer.cs.style_period.České umění 19. st.": "České umenie 19. st.",
+    "importer.cs.style_period.České umění 20. st.": "České umenie 20. st.",
+    "importer.cs.style_period.České umění 21. st.": "České umenie 21. st."
+}
```
eronisko, line 5: I know I'm late to the party, but was there a special reason for creating this JSON lang file? (We don't have any other JSON lang files)

rastislav-chynoransky: I know this is not a perfect solution, but dots in the identifier break translation.
Now I think that we should rewrite importers so that they don't use translations at all...

PR #708, opened 2022-07-14
Laravel 9 upgrade -- part 2 (configs & middleware)

eronisko:
* Config and some other system (e.g. Middleware) files have been updated to their 'skeleton' version, i.e. they are now closer to a clean Laravel project. I've tried to not change any existing behaviour

* lang files have been moved from /resources/lang to /lang (as per new convention)

❌ Tests are expected to fail in this PR


igor-kamil (approved):
💥  this is huge! 
but seems and works OK (even sentry exception is gone)

eronisko:
Thanks for the review! 🙏

PR #709, opened 2022-07-14
Laravel 9 upgrade -- part 3

eronisko:
* Fixing tests
* 
TODO
* Replacement for lorisleiva/laravel-deployer
* Replacement for laravelium/sitemap
* Changelog


.env.testing.example
```diff
@@ -1,13 +1,10 @@
...
\ No newline at end of file
+
+DB_HOST=localhost
+DB_DATABASE=webumenia_test
+DB_USERNAME=webumenia
+DB_PASSWORD=webumenia
+
+ES_PREFIX=webumenia_test
+
+CACHE_DRIVER=redis
```
igor-kamil, line 10: don't we want to keep default (array? nothing in `.env`)
any reason to "predefine" redis? or just to follow default L9 .env? 

```suggestion
```

.github/workflows/develop-autodeploy.yml
```diff
@@ -5,29 +5,25 @@ on:
...
-    - run: php artisan deploy test --branch develop
+      - uses: actions/checkout@v2
+
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: '8.1'
+
+      - name: Deploy
+        uses: deployphp/action@v1
```
igor-kamil, line 24: much better!

.env.testing.example
igor-kamil: now I see. for testing, right?

app/Console/Commands/MakeSitemap.php
```diff
@@ -1,26 +1,28 @@
...
-        'Authority',
-        'Collection',
-        'Item',
+        Article::class,
+        Authority::class,
+        Collection::class,
+        Item::class,
     ];
 
     /**
```
igor-kamil, line 28: [dust] once you are updating this command (thank you for it 🙏 )
what about to align with the namespace and rename `sitemap:make` -> `make:sitemap`?

app/Providers/RouteServiceProvider.php
```diff
@@ -47,5 +47,8 @@ protected function configureRateLimiting()
         RateLimiter::for('api', function (Request $request) {
             return Limit::perMinute(60)->by($request->user()?->id ?: $request->ip());
         });
+        RateLimiter::for('downloads', function (Request $request) {
+            return Limit::perMinute(5)->by($request->user()?->id ?: $request->ip());
```
igor-kamil, line 51: that low?

routes/web.php
```diff
@@ -211,14 +211,14 @@ function()
...
-    Route::get('dielo/{id}/stiahnut', ['middleware' => 'throttle:5,1', function ($id) {
+    Route::get('dielo/{id}/stiahnut', function ($id) {
         $item = Item::findOrFail($id);
         if ($item->images->isEmpty()) {
             abort(404);
         }
 
         return redirect()->route('image.download', ['id' => $item->images->first()->id]);
-    }]);
+    })->middleware('throttle:downloads');
```
igor-kamil, line 221: 👌

igor-kamil (approved):
Looks and works nicely
I didn't notice any issue
Even translated collection work without problem locally (still necessary to investigate what is happening on test)
Looking forward to see this merged! 👍

.env.testing.example
eronisko: Yes, exactly, used for testing. It is not necessary to test with redis, but the closer we can get to production-like environment, the better imho

app/Providers/RouteServiceProvider.php
eronisko: I think this is what it was previously:

https://github.com/SlovakNationalGallery/webumenia.sk/pull/709/files/67df895ce29846741f9d2cd04148c4619da5bad0#diff-193e04aa1705f6f3e484d7efca8f45fe4d19d0ece3b6e6880d3ea070938a1fadL214

https://github.com/GrahamCampbell/Laravel-Throttle#real-examples
I think `5,1` translates to 5 requests with retention of 1 hour

app/Console/Commands/MakeSitemap.php
eronisko: I think `sitemap:make` is fine -- all the other `make` commands are scaffolding generators.

PR #730, opened 2022-08-05
Enable MySQL strict mode

eronisko:
⚠ This PR will likely unleash some unexpected bugs so we'll likely need to do some thorough testing before releasing

I've published this change to test: https://test.webumenia.sk/

* Enables MySQL strict mode
* Expands some columns to better accommodate real data
* Fixes tests to work in strict mode
* Minor change in AbstractHarvester for the `datestamp` field (cc @rastislav-chynoransky )

## Type of change

# Checklist:


database/migrations/2017_05_03_115235_create_article_translations_table.php
```diff
@@ -20,8 +20,8 @@ public function up()
 
             // translatable attributes
             $table->string('title')->default('');
-            $table->text('summary')->default('');
-            $table->text('content')->default('');
+            $table->text('summary');
+            $table->text('content');
```
igor-kamil, line 24: shouldn't it be at least `->nullable()`?

igor-kamil (commented):
right move. but ... I would be very "careful". this might break things way too easy... 
one of the issues is wrong default on every timestamp field `0000-00-00 00:00:00` what in compbination with `'strict' => true` prevents any alter on these tables
in my humble opinion - this should be fixed as well

database/migrations/2017_05_03_115235_create_article_translations_table.php
eronisko: This is actually changed to nullable in 2020_05_07_115534_allow_nulls_in_article_translations.php

eronisko:
Deployed to https://test.webumenia.sk/ again

app/Http/Controllers/ItemController.php
```diff
@@ -135,7 +135,9 @@ protected function processForm(Form $form) {
         $item = $form->getData();
         $form->handleRequest();
 
-        if ($form->isSubmitted() && $form->isValid()) {
+        if ($form->isSubmitted()) {
+            Validator::make(Request::all(), ['item.author' => 'required'])->validate();
```
igor-kamil, line 139: 👌

igor-kamil (approved):
ready to go 🚀

PR #737, opened 2022-08-16
Fix/webumenia 1850 order by iipimg url

rastislav-chynoransky:
Fixes https://jira.sng.sk/browse/WEBUMENIA-1850

## Type of change

## Checklist:

### DIFF
--- a/database/migrations/2022_08_16_225708_drop_unused_columns_from_item_images.php
+++ b/database/migrations/2022_08_16_225708_drop_unused_columns_from_item_images.php
@@ -1,0 +1,35 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration {
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('item_images', function (Blueprint $table) {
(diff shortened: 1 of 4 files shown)


database/migrations/2022_08_16_225708_drop_unused_columns_from_item_images.php
```diff
@@ -0,0 +1,35 @@
...
+     * @return void
+     */
+    public function down()
+    {
+        Schema::table('item_images', function (Blueprint $table) {
+            $table->string('title')->nullable();
+            $table->integer('order')->unsigned();
+        });
+    }
+};
```
eronisko, line 35: Just for the sake of completeness: this is missing the unique index re-creation & the columns would probably not be re-created in the same positions

eronisko (approved):

database/migrations/2022_08_16_225708_drop_unused_columns_from_item_images.php
rastislav-chynoransky: Thanks! Do we bother with reversed data migrations? Or is it enough to simply create the unique index? It would fail if there are data in the table though.

eronisko: Oh, I see now... OK, don't worry about the index then -- I understand it's not being used anyway. LGTM.

PR #743, opened 2022-08-19
Hotfix: disable strict mode temporarily

eronisko:
To prevent outage


igor-kamil (approved):
don't forget `master` &rarr; `develop`

PR #753, opened 2022-08-26
fix: data seeders

FrantisekMichalSebestyen:
# Description

## Type of change

# Checklist:

### DIFF
--- a/database/seeders/HarvestsTableSeeder.php
+++ b/database/seeders/HarvestsTableSeeder.php
@@ -2,6 +2,7 @@
 
 namespace Database\Seeders;
 
+use App\SpiceHarvesterHarvest;
 use Illuminate\Database\Seeder;
 use Illuminate\Support\Facades\Date;
 use Illuminate\Support\Facades\DB;
@@ -15,7 +16,7 @@ class HarvestsTableSeeder extends Seeder
      */
     public function run()
     {
-        $hasrvests = [
+        $harvesters = [
             [
                 'base_url' => 'http://www.webumenia.sk/oai-pmh/',
                 'metadata_prefix' => 'oai_dc',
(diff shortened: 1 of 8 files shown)


eronisko (changes requested):
Thanks for this! It's almost good to go, just a few things I'd like fixed:

* remove the `truncate()` calls -- please remove them from all other seeders too
    * That means we'll be also able to remove the `FOREIGN_KEY_CHECKS` code
* remove all `created_at` and `updated_at` values -- we'll let Eloquent/Laravel handle this
* Please use ` Illuminate\Support\Facades\Date`'s `Date::now()` instead of the `date` function -- we don't have to worry about date format anymore

database/seeders/UsersTableSeeder.php
```diff
@@ -7,51 +7,36 @@
...
                 'email' => '',
                 'password' => Hash::make('press'),
                 'role' => 'editor',
-                'created_at' => $now,
-                'updated_at' => $now
-            ]
+            ],
         ];
 
         DB::table('users')->insert($users);
```
eronisko, line 40: Please use the `User` model instead of `DB` (otherwise you won't get the timestamps filled out)

database/seeders/HarvestsTableSeeder.php
```diff
@@ -29,14 +25,10 @@ public function run()
...
                 'status_messages' => '',
-                'initiated' => $now,
-                'created_at' => $now,
-                'updated_at' => $now
-            ]
+                'initiated' => Date::now(),
+            ],
         ];
 
         DB::table('spice_harvester_harvests')->insert($hasrvests);
```
eronisko, line 32: Please use `SpaceHarvesterHarvest` model instead of `DB`

eronisko (commented):
Please use model classes (or factories) instead of `DB` and this is good to go 👍 Nicely done!

eronisko (approved):
Thank you 👍

PR #754, opened 2022-08-26
Fix Item::related scope to match catalog search query

eronisko:
in order to match catalog search Elastic query

## Type of change

# Checklist:

### DIFF
--- a/app/Item.php
+++ b/app/Item.php
@@ -631,9 +631,12 @@ class Item extends Model implements IndexableModel, TranslatableContract
 
     public function scopeRelated($query, Item $item)
     {
-        return $query->whereTranslation('related_work', $item->related_work)
-            ->where('author', '=', $item->author)
-            ->orderBy('related_work_order');
+        $relatedIds = Item::search()
+            ->where('related_work', $item->related_work)
+            ->whereIn('author', $this->makeArray($item->author))
+            ->keys();
+
+        return $query->whereIn('id', $relatedIds)->orderBy('related_work_order');
     }
 
     public function getAuthorsWithLinks()
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file[^1].
 ## [Unreleased]
 ### Fixed
 - article category on homepage
+- Item::related scope to match catalog search query
 
 ### Added
 - default Scout config to enable Scout search functionality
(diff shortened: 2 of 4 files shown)


app/Item.php
```diff
@@ -616,7 +616,6 @@ public function scopeHasImage($query, $hasImage = true)
     public function scopeRelated($query, Item $item)
     {
         return $query->whereTranslation('related_work', $item->related_work)
-            ->where('author', '=', $item->author)
```
igor-kamil, line 619: this is not a proper fix
first of all... `related_work` might be very vague (e.g. `skicár`) and you wouldn't get the same amount of results in artwork detail and in catalogue view (e.g. https://www.webumenia.sk/katalog?related_work=skicár&author=Novák%2C%20Ján vs https://www.webumenia.sk/katalog?related_work=skicár ).

I think the point is to unify the query to get the same results. Hence I suggest to follow  https://github.com/SlovakNationalGallery/webumenia.sk/blob/develop/resources/views/dielo.blade.php#L315-L316 (which gets better results - since it's using only `first_author`)

And fix it with something like
```
        return $query->whereTranslation('related_work', $item->related_work)
            ->where('author', 'like', $item->first_author . '%')
            ->orderBy('related_work_order');
```
or 
```
        return $query->whereTranslation('related_work', $item->related_work)
            ->where('author', 'like',  '%' . $item->first_author . '%')
            ->orderBy('related_work_order');
```

igor-kamil (changes requested):
see comment

app/Item.php
eronisko: Thanks for the pointers. I've tried making this into a `LIKE` query just now and while it works, it always takes over a second to run.

Not sure if this is the right avenue now, seems like a job for ES in this case (the zoom controller is the only place using this query anyway)

eronisko: Re-worked this so that it does an ES search before retrieving items from DB

eronisko:
@igor-kamil @FrantisekMichalSebestyen @rastislav-chynoransky This is finally passing tests, so ready for another review :)

app/Item.php
```diff
@@ -615,9 +615,12 @@ public function scopeHasImage($query, $hasImage = true)
 
     public function scopeRelated($query, Item $item)
     {
-        return $query->whereTranslation('related_work', $item->related_work)
-            ->where('author', '=', $item->author)
-            ->orderBy('related_work_order');
+        $relatedIds = Item::search()
+            ->where('related_work', $item->related_work)
+            ->whereIn('author', $this->makeArray($item->author))
+            ->keys();
```
igor-kamil, line 621: 🙌

igor-kamil (approved):
LGTM

PR #757, opened 2022-08-31
Webumenia 1527 add artwork quote to license modal

FrantisekMichalSebestyen:
# Description
Add artwork citation (same format as "copy citation") to license modal (displayed after clicking "download artwork")
Snímka obrazovky 2022-08-31 o 10 16 11

Fixes [https://jira.sng.sk/browse/WEBUMENIA-1527)]

## Type of change

# Checklist:

### DIFF
--- a/app/Item.php
+++ b/app/Item.php
@@ -172,7 +172,7 @@ class Item extends Model implements IndexableModel, TranslatableContract
 
     public function getCitation()
     {
-        return $this->author . ' - ' . $this->title . ', ' . $this->getDatingFormated() . ', ' . $this->gallery . ', ' . $this->getUrl();
+        return $this->getTitleWithAuthors() . ', ' . $this->getDatingFormated() . ', ' . $this->gallery . ', ' . $this->getUrl();
     }
 
     public static function loadValidatorMetadata(ClassMetadata $metadata) {
--- a/app/Item.php
+++ b/app/Item.php
@@ -170,6 +170,11 @@ class Item extends Model implements IndexableModel, TranslatableContract
 
     protected $useTranslationFallback;
 
+    public function getCitation()
+    {
+        return $this->author . ' - ' . $this->title . ', ' . $this->getDatingFormated() . ', ' . $this->gallery . ', ' . $this->getUrl();
+    }
+
     public static function loadValidatorMetadata(ClassMetadata $metadata) {
         $metadata->addGetterConstraint('images', new Valid());
     }
--- a/public/js/components/share_buttons.js
+++ b/public/js/components/share_buttons.js
@@ -1,10 +1,8 @@
-function copyToClipboard(url, message, el) {
-    var $temp = $("<input>");
-    $("#shareLink").append($temp);
-    $temp.val(url).select();
-    var successful = document.execCommand("copy");
-    $temp.remove();
-    var msg = successful ? message : 'Whoops, not copied...';
+async function copyToClipboard(url, message, el) {
+    var msg = 'Whoops, not copied...';
+    await navigator.clipboard.writeText(url).then(() => {
+      msg = message;
+    });
     el.attr('data-original-title', msg).tooltip('show');
     setTimeout(function () {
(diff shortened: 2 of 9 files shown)


FrantisekMichalSebestyen:
There is a weird space between author title and year

resources/views/dielo.blade.php
```diff
@@ -473,14 +477,12 @@ class="fa fa-download"></i>
...
-                'author' => $item->author,
-                'title' => $item->title,
-                'date' => $item->getDatingFormated(),
-                'owner' => $item->gallery,
-                ] ) !!}
+                {!! trans('dielo.modal_license_body-header') !!}
+                <div>
+                    <pre><code>{{ $citation }}</code></pre>
+                </div>
+                @include('components.copy_button', ['text' => $citation])
```
FrantisekMichalSebestyen, line 484: I will need some guidance on where to put copy button

public/js/components/share_buttons.js
```diff
@@ -1,10 +1,8 @@
-function copyToClipboard(url, message, el) {
-    var $temp = $("<input>");
-    $("#shareLink").append($temp);
-    $temp.val(url).select();
-    var successful = document.execCommand("copy");
-    $temp.remove();
-    var msg = successful ? message : 'Whoops, not copied...';
+async function copyToClipboard(url, message, el) {
```
FrantisekMichalSebestyen, line 1: I was trying to keep the OG way of doing this, however it wasn't working. Hopefully this is ok

resources/views/dielo.blade.php
FrantisekMichalSebestyen: I will also need some guidance on documentation

igor-kamil: it's fine
just add some space under
Snímka obrazovky 2022-09-21 o 14 43 56

resources/views/dielo.blade.php
```diff
@@ -40,6 +40,10 @@
 </section>
 @endif
 
+@php
+    $citation = $item->author . ' - ' . $item->title . ', ' . $item->getDatingFormated() . ', ' . $item->gallery . ', ' . $item->getUrl();
```
igor-kamil, line 44: well.. I am not an expert in formating quotations. is this 100% right?

I would prefer maybe something like:
```
$citation = $item->getTitleWithAuthors().' (' . $item->getDatingFormated() . ")\n" . $item->gallery . "\n" . $item->getUrl();
```

which returns
Snímka obrazovky 2022-09-22 o 11 29 51

@ustengrav ? 

instead of using `@php` within view template... I would consider using "virtual attribute/getter" https://laravel.com/docs/9.x/eloquent-mutators#defining-an-accessor - e.g. in model: `getCitationAttribute()` and  then used in the view: `$item->citation`

igor-kamil (commented):
looks good. just fix the spacing. the format of the citation can be changed anytime later....

resources/views/dielo.blade.php
FrantisekMichalSebestyen: ur right @php tag is not good, however not sure about need to use acessors:

> An accessor transforms an Eloquent attribute value when it is accessed

I do not have eloquent attribute value yet, IMHO isn't function in model enough ?

app/Item.php
```diff
@@ -170,6 +170,11 @@ class Item extends Model implements IndexableModel, TranslatableContract
 
     protected $useTranslationFallback;
 
+    public function getCitation()
+    {
+        return $this->author . ' - ' . $this->title . ', ' . $this->getDatingFormated() . ', ' . $this->gallery . ', ' . $this->getUrl();
```
DigitalArtsAFAD, line 175:
```suggestion
        return $item->getTitleWithAuthors() . ', ' . $this->getDatingFormated() . ', ' . $this->gallery . ', ' . $this->getUrl();
```

DigitalArtsAFAD (approved):
🚀

PR #763, opened 2022-09-14
feat: parse article from admin to display article-thumbnail

FrantisekMichalSebestyen:
Admin view:
Snímka obrazovky 2022-09-14 o 11 36 00

Article with thumbnail:
Snímka obrazovky 2022-09-14 o 11 35 49


app/Http/Controllers/ClanokController.php
```diff
@@ -98,6 +105,7 @@ public function getDetail($slug)
         }
         $article->view_count += 1;
         $article->save();
+        $article->content = Blade::render($this->replaceTags("$article->content"));
```
igor-kamil, line 108: what do you @eronisko think about this? 
probably it's ok - since we have control of the article content (it's not from unknown users)

however I would consider something like:
```
        $html = Str::of($html)->replaceMatches('/[ article_thumbnail id="(.*)" \/]/', function ($match) {
                if (empty($match[1])) return '';
​
                $article = Article::find($match[1]);
                $article_thumbnail = view('components.x-article_thumbnail')
                   ->with('media', $article)
                   ->render();
                return $article_thumbnail;
            });
```

(proposed code is not tested... just wrote here)

ta avoid launching any unexpected php code from the content

public/ckeditor/plugins/thumbnail/lang/cs.js
```diff
@@ -0,0 +1,8 @@
+'use strict';
+
+(function (CKEDITOR) {
+    CKEDITOR.plugins.setLang('thumbnail', 'cs', {
+        title: 'Ukázka',
+        idRequired: 'Prosím zadefinujte id ukazky (číslo)',
```
igor-kamil, line 6:
```suggestion
        idRequired: 'Prosím zadefinujte id článku (číslo)',
```

public/ckeditor/plugins/thumbnail/lang/sk.js
```diff
@@ -0,0 +1,8 @@
+'use strict';
+
+(function (CKEDITOR) {
+    CKEDITOR.plugins.setLang('thumbnail', 'sk', {
+        title: 'Ukážka',
+        idRequired: 'Prosím zadefinujte id ukážky (číslo)',
```
igor-kamil, line 6:
```suggestion
        idRequired: 'Prosím zadefinujte id článku (číslo)',
```

igor-kamil (commented):
looks nice. since this is somehow defining the way we are going to use components within article/collection content - it's good to design it well. 
I think you can turn PR from draft to regular PR and I would like to see CR from @eronisko

app/Http/Controllers/ClanokController.php
rastislav-chynoransky: Yes, please, don't interpret the user input. That might quite easily cause errors (since blade syntax could be part of the text).

eronisko (changes requested):
First of all sorry for responding to this so late. I actually had a comment drafted up for 7 days but haven't been able to finish it until now.

I think the approach with square brackets is sound, but I'd defer from creating any generic solutions at all. I have two possible variants in mind

## Tag with an ID
Similar to what is proposed here, I'd suggest a syntax like this:
```
[ArticleTeaser id=123]
```

Where `id` would be the only supported attribute and I'd parse it like this:
1. Look for something like `[ArticleTeaser .*]` using regex
2. Extract only the supported attributes (`id`)
3. Turn it into a Blade expression

A good tip might be replacing `[ ]` with `< />` (similar to the proposed solution) and then using a HTML parser to extract the attributes.

## Simple tag, no attributes
The disadvantage of the solution above is that at some point we'd have to query the database from a Blade component, which is not a good practice.

If all we need is a teaser to a single article, we can define an "teaser article" attribute on the main object within the admin interface (e.g. `teaser_article_id`) and all we have to do then is to insert `[ArticleTeaser]` somewhere in the text. So the custom block would just serve as a "HERE" pointer for the rendering mechanism, whereas all the data querying would happen outside.

Another advantage of this approach is that we can implement some outside logic for deciding which teasers to show. A disadvantage would be perhaps less flexibility for each article.

resources/views/components/article_thumbnail.blade.php
```diff
@@ -1,9 +1,21 @@
...
+if (isset($id)) {
+    $article = App\Article::where('publish', '=', 1)
+        ->where('id', $id)
+        ->firstOrFail();
+}
+
+$url = $article->getUrl();
+
+@endphp
+
```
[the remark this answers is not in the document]
FrantisekMichalSebestyen: You're right personally, I would make two separate components. If it's alright. The reasoning behind it is that the logic (things supposed to be displayed, what is clickable) changes between these two and I do not expect that the wrapper will be reused on some other page.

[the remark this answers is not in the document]
eronisko: Cool, sounds good :)

igor-kamil:
I vote for "_Tag with an ID_" as @eronisko proposed. 
I see it as the most flexible option - and also quick to implement.

FrantisekMichalSebestyen:
@eronisko @igor-kamil  please review the proposed solution - "Tag with id". Keep in mind that style article teaser is in another PR

resources/views/frontend/articles/show.blade.php
```diff
@@ -108,7 +108,7 @@
             </div>
             <div class="row">
                 <div class="col-md-8 col-md-push-2 long-text">
-                    {!! $article->content !!}
+                    {!! $article->getParsedContent() !!}
                 </div>
```
igor-kamil, line 112: In case you renamed the method to `getParsedContentAttribute()` 
you can use:

```suggestion
                    {!! $article->parsed_content !!}
                </div>
```

app/Article.php
```diff
@@ -141,4 +143,19 @@ public function getReadingTimeAttribute()
     {
         return getEstimatedReadingTime($this->summary . ' ' . $this->content, \App::getLocale());
     }
+
+    public function getParsedContent()
+    {
+        $content = Str::of($this->content)->replaceMatches("/\[x-article_teaser id=&#39;.*&#39;\/\]/", function ($match) {
+            if (empty($match[0])) return '';
+            $id = Str::of($match[0])->between("id=&#39;","&#39;");
+            $article = Article::find($id);
```
igor-kamil, line 152: just put id into parenthesis in the regexp above and use

```suggestion
            $article = Article::find($match[1]);
```

igor-kamil (commented):
it looks good! almost there 👍 
I have few comments, but I don't see any real blocker from my point of view

app/Article.php
eronisko: It should also only filter published articles (please add a scope)

public/ckeditor/plugins/articleTeaser/lang/sk.js
```diff
@@ -0,0 +1,8 @@
+'use strict'
+
+;(function (CKEDITOR) {
+    CKEDITOR.plugins.setLang('articleTeaser', 'sk', {
+        title: 'Ukážka',
+        idRequired: 'Prosím zadefinujte id článku (číslo)',
```
eronisko, line 6:
```suggestion
        idRequired: 'Zadajte id článku (číslo)',
```

eronisko (changes requested):
The regexes need a bit more work, but we're getting there :)

> Keep in mind that style article teaser is in another PR

If this PR doesn't deliver a full feature, then please don't merge it into a develop -- make a 'feature' branch instead. Merge your different PRs into that and then we'll merge it into develop all at once.

Whatever is merged to develop should be ready to go to prod at any time.

app/Article.php
```diff
@@ -141,4 +143,19 @@ public function getReadingTimeAttribute()
...
+    public function getParsedContentAttribute()
+    {
+        $content = Str::of(html_entity_decode($this->content))->replaceMatches("/\[article_teaser id=(.*?)\]/", function ($match) {
+            if (empty($match[1])) return '';
+            $id = Str::of($match[1]);
+            $article = Article::published()->find($id);
+
+            if(!$article) return '';
+            
+            $article_teaser = Blade::render('<x-article_teaser :article="$article" :url="$url" />', ['article' => $article, 'url' => $article->getUrl()]);
```
eronisko, line 156:
```suggestion
            $article_teaser = Blade::render('<x-article_teaser :article="$article" />', ['article' => $article]);
```

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ All notable changes to this project will be documented in this file[^1].
 ## [2.68.0] - 2022-10-13
 ### Added
 - default Scout config to enable Scout search functionality
+- parse article teaser from admin
```
eronisko, line 11: This should come under 'Unreleased'

eronisko (approved):
LGTM, thank you :)

PR #766, opened 2022-09-19
style: article thumbnail component

FrantisekMichalSebestyen:
# Description

Style article thumbnail component inside list
I will follow article thumbnail component inside article in another PR

SM: 
Snímka obrazovky 2022-09-19 o 10 52 19

MD:
Snímka obrazovky 2022-09-19 o 10 52 47

 
LG:
Snímka obrazovky 2022-09-19 o 10 52 02

## Type of change

# Checklist:


resources/views/components/article_thumbnail_index.blade.php
```diff
@@ -0,0 +1,42 @@
+@php
+
+$url = $url ?? $article->getUrl();
+$showEduTags = $showEduTags ?? false;
+
+@endphp
+<div class="tailwind-rules">
+    <div
+        class="tw-container tw-mx-auto tw-flex tw-h-[475px] tw-max-w-screen-xl tw-flex-wrap tw-px-6 tw-py-4 md:tw-h-[400px] md:tw-flex-nowrap md:tw-py-8">
```
eronisko, line 9: Please specify the custom heights in `rem` rather than `px`. 

(Almost everything in Tailwind is in rem, e.g. https://tailwindcss.com/docs/customizing-spacing#default-spacing-scale)

resources/views/components/article_thumbnail_index.blade.php
```diff
@@ -0,0 +1,42 @@
+@php
+
+$url = $url ?? $article->getUrl();
+$showEduTags = $showEduTags ?? false;
```
eronisko, line 4: Please turn these into `@props`: https://laravel.com/docs/9.x/blade#data-properties-attributes

`$article` should be included in the `@props` list too, that way we should get errors earlier.

resources/views/components/article_thumbnail_index.blade.php
```diff
@@ -0,0 +1,42 @@
...
+
+@endphp
+<div class="tailwind-rules">
+    <div
+        class="tw-container tw-mx-auto tw-flex tw-h-[475px] tw-max-w-screen-xl tw-flex-wrap tw-px-6 tw-py-4 md:tw-h-[400px] md:tw-flex-nowrap md:tw-py-8">
+        <div class="tw-h-1/3 tw-w-full md:tw-mr-4 md:tw-h-full md:tw-w-1/2">
+            @if ($article->main_image)
+                <a href="{{ $url }}" class="tw-h-full">
+                    <img src="{!! $article->getThumbnailImage() !!}"
```
eronisko, line 13: As long as `getThumbnailImage` is returning a non-html string, we don't have to use the `{!! !!}` brackets
```suggestion
                    <img src="{{ $article->getThumbnailImage() }}"
```

resources/views/components/article_thumbnail_index.blade.php
```diff
@@ -0,0 +1,42 @@
...
+            @endif
+            <div class="tw-pb-2 tw-text-2xl tw-font-semibold">
+                <a href="{{ $url }}">
+                    {{ $article->title }}
+                </a>
+            </div>
+            <div class="tw-pb-4 tw-text-base tw-text-gray-600">
+                <span class="hover:tw-underline">
+                    {{ $article->author }}</span> &nbsp;&middot;&nbsp;
+                {!! $article->created_at->format('d. m. Y') !!}
```
eronisko, line 32: I'd use the `@date` or `@dateShort` directive (from AppServiceProvider) here, to keep the date format consistent with other places.

(Also no need for `{!!`)

resources/views/frontend/articles/index.blade.php
```diff
@@ -49,21 +49,16 @@
...
-                    ])
-                </div>
-                @if ($i % 2 == 1)
-                    <div class="clearfix"></div>
-                @endif
-            @endforeach
-        </div>
-        <div class="row text-center">
+    </div>
+    <div class="tailwind-rules">
```
eronisko, line 53: Do we need this wrapping div here?

resources/views/components/article_thumbnail_index.blade.php
```diff
@@ -0,0 +1,42 @@
+@php
```
eronisko, line 1: I'd suggest re-using the `article_thumbnail` name, especially since we're deleting the old component. The `_index` suffix does not add any more information IMHO.

eronisko (changes requested):
Thanks. Code-wise this looks good, except for a few notes, mostly for code carried over from the old component.

I'm not sure whether the design will work well with the data we have today (I don't think the current thumbnail resolution will be sufficient). But that would be for a discussion elsewhere.

resources/views/components/article_thumbnail_index.blade.php
igor-kamil: I think I see the reason (because there is going to be another article_thumbnail inside article)
However I prefer to keep it consistent with previous code... please don't rename (until really necessary)

resources/views/components/article_thumbnail_index.blade.php
eronisko, line 13: Please take a look at `HasHeaderImage#getThumbnailImage` -- we might have to change the resolution for thumbnails if we want them larger

resources/views/components/article_thumbnail_index.blade.php
eronisko: That's not necessary in my opinion (let's keep the PR lean)

resources/views/components/article_thumbnail_index.blade.php
FrantisekMichalSebestyen: Yes, true. Also, we should note somewhere, that we will need to rerender old article thumbnails at some point.

FrantisekMichalSebestyen:
> from my point of view... there are two (visual) issues. first is hard-cutting the overflowed text Snímka obrazovky 2022-09-21 o 11 56 53
> 
> try to set height to avoid cutting the line (and test it for all breakpoints) - I think it might be possible second option is to apply "fade-out" with gradient (similar what used to seen on paywalls)
> 
> second issue is spacing. it simply doesn't follow the proposed design: 190982660-3d50d325-4a5b-4c73-9f9c-c251cdbf43d3
> 
> vs
> 
> Snímka obrazovky 2022-09-22 o 11 00 14

I was thinking about this and there are simply too much variables: 
1. Category is either there or not 
2. Title is from 1 to 3 lines
3. and ofc the breakpoints
I pushed the fade out but will consult @memb

resources/views/components/article_thumbnail.blade.php
```diff
@@ -1,35 +1,44 @@
-@php
+@props([
+    'article' => null,
```
eronisko, line 2:
```suggestion
    'article',
```

resources/views/components/article_thumbnail.blade.php
```diff
@@ -1,35 +1,44 @@
...
+    'url' => $article->getUrl(),
```
eronisko, line 3: Does this work? (It seems that we're defining and using the`$article` variable in one statement, so it's an honest question)

A good test would be to re-name the `article` prop to something else

app/Concerns/HasHeaderImage.php
```diff
@@ -136,7 +136,7 @@ public function getThumbnailImage($full = false)
 
         if (!file_exists($preview_full_path)) {
             try {
-                \Image::make($full_path)->fit(600, 250)->save($preview_full_path);
+                \Image::make($full_path)->fit(1200, 480)->save($preview_full_path);
```
eronisko, line 139: On a second thought, we'll probably need this for the mobile layout, so please roll this back to 600+250 -- sorry

resources/views/components/article_thumbnail.blade.php
```diff
@@ -1,35 +1,44 @@
...
-  </div>
-  @endif
-  <p class="meta mt-3">{!!$article->published_date!!} / {!!$article->author!!}</p>
+<div class="tailwind-rules">
+    <div
+        class="tw-container tw-mx-auto tw-flex tw-h-[30rem] tw-max-w-screen-xl tw-flex-wrap tw-px-6 tw-py-4 md:tw-h-[25.75rem] md:tw-flex-nowrap">
+        <div class="tw-h-1/3 tw-w-full md:tw-h-full md:tw-w-1/2">
+            @if ($article->main_image)
+                <a href="{{ $url }}" class="tw-h-full">
+                    <img src="{{ $article->getThumbnailImage() }}"
```
eronisko, line 13: We should incorporate the thumbnail image size into a [responsive image](https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images) attribute (`srcset`)

Please either include the thumbnail size in `HasHeaderImage#getHeaderImageSrcsetAttribute` or suggest a new method to generated `srcset` for the thumbnail `img` tag.

eronisko (commented):
Code-wise this looks fine (just please render the "thumbnail" image as a responsive image)

In terms of design, I'm not sure. Still think think this typography is a bit too generous:
image

eronisko:
I've deployed the PR to test:

```
 dep deploy test --branch WEBUMENIA-1558-article-thumbnail-component
```

@ustengrav @igor-kamil , can you please take a look?

FrantisekMichalSebestyen:
@eronisko Sorry you are right about the typograhpy on mobile I somehow missed it. Fixed it in new commit

resources/views/components/article_thumbnail.blade.php
FrantisekMichalSebestyen: It should. We are not passing url attribute in `_frontend/articles/index.blade.php_ `, but we are able to click on image to get in the article.

eronisko (commented):
image

The author is underlined on hover, but it's not a link. Not sure what the desired state is.

resources/views/components/article_thumbnail.blade.php
```diff
@@ -1,35 +1,44 @@
...
+            @endif
+        </div>
+        <div
+            class="tw-flex tw-h-2/3 tw-w-full tw-flex-col tw-pt-4 md:tw-box-border md:tw-h-full md:tw-max-h-full md:tw-w-1/2 md:tw-px-12 md:tw-py-10">
+            @if ($article->category?->name)
+                <div class="tw-pb-3 tw-text-sm tw-font-medium tw-text-gray-600 md:tw-text-lg">
+                    {{ $article->category->name }}
+                </div>
+            @endif
+            <div class="text-2xl tw-pb-2 tw-font-semibold md:tw-text-4xl">
```
eronisko, line 25: Please take a look at the `text-2xl` class (it should be `tw-text-2x`)

FrantisekMichalSebestyen: I will ask memb about the underline

resources/views/components/article_thumbnail.blade.php
```diff
@@ -1,35 +1,44 @@
...
+                <a href="{{ $url }}" class="tw-h-full">
+                    <img src="{{ $article->getThumbnailImage() }}"
+                        class="tw-h-full tw-w-full tw-object-cover" alt="{{ $article->title }}">
+                </a>
+            @endif
+        </div>
+        <div
+            class="tw-flex tw-h-2/3 tw-w-full tw-flex-col tw-pt-4 md:tw-box-border md:tw-h-full md:tw-max-h-full md:tw-w-1/2 md:tw-px-12 md:tw-py-10">
+            @if ($article->category?->name)
+                <div class="tw-pb-3 tw-text-sm tw-font-medium tw-text-gray-600 md:tw-text-lg">
```
eronisko, line 21: Please add `capitalize`

FrantisekMichalSebestyen: Are you sure ? We don't do it anywhere else

FrantisekMichalSebestyen: Can it be that data in test databse are different than the ones on production ? therefore category is lowercase

eronisko: We're capitalizing first letter here https://github.com/SlovakNationalGallery/webumenia.sk/blob/develop/resources/views/home/index.blade.php#L278 for example. But using the CSS class is more elegant IMHO

resources/views/components/article_thumbnail.blade.php
```diff
@@ -1,16 +1,12 @@
...
+@props(['article', 'url' => $article->getUrl(), 'showEduTags' => false])
 
 <div class="tailwind-rules">
     <div
         class="tw-container tw-mx-auto tw-flex tw-h-[30rem] tw-max-w-screen-xl tw-flex-wrap tw-px-6 tw-py-4 md:tw-h-[25.75rem] md:tw-flex-nowrap">
         <div class="tw-h-1/3 tw-w-full md:tw-h-full md:tw-w-1/2">
             @if ($article->main_image)
                 <a href="{{ $url }}" class="tw-h-full">
-                    <img src="{{ $article->getThumbnailImage() }}"
+                    <img srcset="{{ $article->getHeaderImageSrcsetAttribute() }}"
```
FrantisekMichalSebestyen, line 9: this should be enough right ? @eronisko since the width of image is determined by flexbox

resources/views/components/article_thumbnail.blade.php
FrantisekMichalSebestyen: tw-capitalize capitalizes first char of every letter, therefore i used Str::ucfirst

resources/views/components/article_thumbnail.blade.php
```diff
@@ -1,35 +1,44 @@
...
-  <p class="meta mt-3">{!!$article->published_date!!} / {!!$article->author!!}</p>
+<div class="tailwind-rules">
+    <div
+        class="tw-container tw-mx-auto tw-flex tw-h-[30rem] tw-max-w-screen-xl tw-flex-wrap tw-px-6 tw-py-4 md:tw-h-[25.75rem] md:tw-flex-nowrap">
+        <div class="tw-h-1/3 tw-w-full md:tw-h-full md:tw-w-1/2">
+            @if ($article->main_image)
+                <a href="{{ $url }}">
+                    <img srcset="{{ $article->getHeaderImageSrcsetAttribute() }}"
+                        src="{{ $article->getThumbnailImage() }}"
+                        sizes="(max-width: 768w) 100vw, 50vw"
```
eronisko, line 11: The thumbnails are kinda blurry at ~1024px width. 

Compare 

```
sizes="(max-width: 768w) 100vw, 50vw"
```
image

to

```
sizes="(max-width: 768w) 100vw, 150vw"
```
image

eronisko (changes requested):
I don't mean to stand in the way of progress. Code-wise, I think this looks good. The design (implementation?) needs still needs some work here. 

This still looks broken:
image

I offer https://tailwindcss.com/blog/multi-line-truncation-with-tailwindcss-line-clamp up for consideration.

At certain resolutions, this just doesn't look very good:

### 768px
image

### ~700px
(this I could live with, but wanted to point out)
image

### 320px
keeps cutting off last line
image

Feel free to reach out to myself, @igor-kamil or @ustengrav, if you want to discuss how to solve some of these issues.

resources/views/components/article_thumbnail.blade.php
eronisko: Note that yes, the image is only ~50% of the screen, but because its height (not width) is specified here, sometimes it needs to stretch beyond 100% of screen size

eronisko:
Since don't have a home for this yet, I'd suggest closing this PR (but leaving the branch as is)

PR #769, opened 2022-09-22
[item] fix lazy loading for reindex

rastislav-chynoransky:
Removed cross join since it was fetching all items at once

### DIFF
--- a/app/Elasticsearch/Repositories/ItemRepository.php
+++ b/app/Elasticsearch/Repositories/ItemRepository.php
@@ -466,6 +466,7 @@ class ItemRepository extends TranslatableRepository
 
                 return $operations;
             })
+            // chunk size = 2 operations * number of locales * 200 items
             ->chunk(2 * count($this->locales) * 200)
             ->each(function ($operations) use (&$processedCount) {
                 $this->elasticsearch->bulk(['body' => $operations]);
--- a/app/Elasticsearch/Repositories/ItemRepository.php
+++ b/app/Elasticsearch/Repositories/ItemRepository.php
@@ -445,27 +445,28 @@ class ItemRepository extends TranslatableRepository
 
         $query
             ->lazy()
-            ->tap(function () use (&$processedCount) {
+            ->tapEach(function () use (&$processedCount) {
                 $processedCount++;
             })
-            ->crossJoin(collect($this->locales))
-            ->flatMap(function ($item_locale) {
-                [$item, $locale] = $item_locale;
+            ->flatMap(function (Item $item) {
+                $operations = [];
 
-                return [
+                foreach ($this->locales as $locale) {
                     // Action
-                    [
+                    $operations[] = [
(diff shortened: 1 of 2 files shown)


app/Elasticsearch/Repositories/ItemRepository.php
```diff
@@ -445,27 +445,28 @@ public function reindexAllLocales(Carbon $viewedSince = null): int
 
         $query
             ->lazy()
-            ->tap(function () use (&$processedCount) {
+            ->tapEach(function () use (&$processedCount) {
```
eronisko, line 448: 👍

app/Elasticsearch/Repositories/ItemRepository.php
```diff
@@ -445,27 +445,28 @@ public function reindexAllLocales(Carbon $viewedSince = null): int
...
+                    ];
 
                     // Data
-                    $item->getIndexedData($locale),
-                ];
+                    $operations[] = $item->getIndexedData($locale);
+                }
+
+                return $operations;
             })
```
eronisko, line 468: Doesn't this do the same thing as before?

app/Elasticsearch/Repositories/ItemRepository.php
```diff
@@ -445,27 +445,28 @@ public function reindexAllLocales(Carbon $viewedSince = null): int
...
-            ->chunk(1000)
+            ->chunk(2 * count($this->locales) * 200)
```
eronisko, line 469: Oh, did this split the actions and data?

eronisko (commented):
What was the problem/reason for this change?

app/Elasticsearch/Repositories/ItemRepository.php
rastislav-chynoransky: No, it just has to be divisible by 2, but this seemed to me more readable what the number means

app/Elasticsearch/Repositories/ItemRepository.php
rastislav-chynoransky: It removes cross join since it was fetching all items at once to build it

eronisko: Oh, I see 😬 thank you

app/Elasticsearch/Repositories/ItemRepository.php
eronisko: Could you please add a comment explaining the formula? (It was not obvious to me)

eronisko (approved):
Thank you for this. I believe this'll also fix https://jira.sng.sk/browse/WEBUMENIA-1887

rastislav-chynoransky:
> Thank you for this. I believe this'll also fix https://jira.sng.sk/browse/WEBUMENIA-1887

Oh, I didn't know about that one. Yes, I think you can close it.

PR #781, opened 2022-10-24
[import] use flysystem storages

rastislav-chynoransky:
CSV files are being moved to common storage with JP2 images.
For now that only applies for PnpKarasekImporter, however in the future there should be more of them.
https://jira.sng.sk/browse/PNP-73


.env.example
```diff
@@ -20,7 +20,7 @@ MAIL_PASSWORD=null
 CACHE_DRIVER=redis
 QUEUE_CONNECTION=database
 
-IIP_BASE_PATH=/iip_base_path
+DG_PUBLIC_IS_PATH=
```
eronisko, line 23: Could you please pick a more verbose name for this? I don't know that `DG` or `IS` is

config/filesystems.php
```diff
@@ -28,6 +28,12 @@
     */
 
     'disks' => [
+        'DG_PUBLIC_IS' => [
```
eronisko, line 31: Same as above -- can you please name this more generously? :)

database/migrations/2022_10_25_104407_add_disk_to_imports_table.php
```diff
@@ -0,0 +1,31 @@
...
+return new class extends Migration {
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('imports', function (Blueprint $table) {
+            $table->string('disk')->default('import');
```
eronisko, line 16: Here I would strongly prefer setting the default in either code or a config file.

tests/Importers/WebumeniaMgImporterTest.php
```diff
@@ -180,11 +182,17 @@ protected function importSingle(array $data)
 
     protected function fakeData(array $data = [])
     {
+        $storage = Storage::fake('DG_PUBLIC_IS');
+        $storage->put('rada_s000123-lomeni_s.jp2', '');
```
eronisko, line 186: It's not obvious to me why & when this file needs to exist

eronisko (changes requested):
This is a nice improvement, and I'm looking forward to it. I have some comments mostly to do with naming, but I'm also not able to run the tests locally. I get this message:

```
  Unable to read file from location: .
```

Can you please help me debug or (better) make it so that I don't have to add/change anything in order to run the tests locally? (Currently the tests are working for me against `develop` branch)

database/migrations/2022_10_25_083719_cleanup_import_records_table.php
```diff
@@ -0,0 +1,39 @@
...
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('import_records', function (Blueprint $table) {
+            $table
+                ->integer('imported_items')
+                ->default(0)
```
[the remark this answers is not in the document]
rastislav-chynoransky: Here I wanted to align it with the other columns `imported_iip` and `wrong_items`. And since the columns are not nullable it seemed like reasonable default for the schema. I'm also ok with creating `ImportRecordFactory` for this purpose, but I find it unnecessary for such primitive defaults.

app/Importers/PnpKarasekImporter.php
```diff
@@ -47,27 +49,38 @@ protected function init()
         };
     }
 
-    public function import(Import $import, SplFileInfo $file)
+    public function import(Import $import, SplFileInfo $file): array
```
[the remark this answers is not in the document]
rastislav-chynoransky: Not sure if I get you right, but the overriding definition must follow that one of the parent method.

config/filesystems.php
rastislav-chynoransky: The name exactly refers to the name of the folder on `imi` server. I was considering naming it `imi` disk, however we're not mounting/referencing the whole `imi` server, only this folder. I think it best describes what it is but I'm open to your suggestions. @igor-kamil had the same comment.

tests/Importers/WebumeniaMgImporterTest.php
rastislav-chynoransky: In `WebumeniaMgImporter` we're importing only those items that have some associated JP2 images therefore the dependency on filesystem.

database/migrations/2022_10_25_083719_cleanup_import_records_table.php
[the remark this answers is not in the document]
eronisko: OK. If other column defaults are set in DB too then I guess it's better to be consistent.

I'm just not a fan of setting non-trivial defaults (anything other than e.g. `0`, or `false`) in DB because it's hard to find them later and any changes will require more DB migrations.

FYI, I don't think it would be necessary to reach for a factory right away, there's a way to [set attribute defaults in Laravel](https://laravel.com/docs/9.x/eloquent#default-attribute-values)

app/Importers/PnpKarasekImporter.php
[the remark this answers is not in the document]
eronisko: I probably asked in the wrong place. Is there much added value in defining the return type in `AbstractImporter::import`?
Can't your IDE/editor figure out what the type will be?

Here's an example without annotation
image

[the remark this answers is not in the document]
rastislav-chynoransky: I think it's the best practice to use typed definitions, but I'm open to discussion.

database/migrations/2022_10_25_083719_cleanup_import_records_table.php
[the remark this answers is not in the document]
rastislav-chynoransky: I didn't know about them, thanks!

config/filesystems.php
eronisko: Well, I don't know what the `imi` server is either :D How about something like `JP2_IMAGES_DIR`? Is it the same as the IIP server? Then maybe `IIP_IMAGES_DIR`?

eronisko: Or even IMPORT_IMAGES_DIR... the fewerabbreviations the better.

rastislav-chynoransky: That’s the thing that we’re moving csv and jpeg files there as well. We can still call it `imi` disk (as is the name of the mounted server), but I’ve mentioned why I didn’t so.

storage/framework/testing/.gitignore
```diff
@@ -0,0 +1,2 @@
+*
+!.gitignore
```
eronisko, line 2: This is probably no longer required

eronisko (approved):
Looks good to me and tests are passing locally no problem.

Thank you very much for all the changes. 

🚀

storage/framework/testing/.gitignore
rastislav-chynoransky: Not required but it follows the current framework template so I'd keep it there.

PR #785, opened 2022-11-04
[import] fix form file list

rastislav-chynoransky:
https://jira.sng.sk/browse/PNP-73

### DIFF
--- a/database/factories/ImportFactory.php
+++ b/database/factories/ImportFactory.php
@@ -2,12 +2,6 @@
 
 namespace Database\Factories;
 
-use App\Importers\GmuRnlImporter;
-use App\Importers\MgImporter;
-use App\Importers\OglImporter;
-use App\Importers\PnpImporter;
-use App\Importers\PnpKarasekImporter;
-use App\Importers\PnpTrienaleImporter;
 use App\Importers\WebumeniaMgImporter;
 use Illuminate\Database\Eloquent\Factories\Factory;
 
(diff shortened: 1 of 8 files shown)


database/factories/ImportFactory.php
```diff
@@ -0,0 +1,39 @@
...
+            'name' => fake()->text(),
+            'class_name' => fake()->randomElement([
+                GmuRnlImporter::class,
+                MgImporter::class,
+                OglImporter::class,
+                PnpImporter::class,
+                PnpKarasekImporter::class,
+                PnpTrienaleImporter::class,
+                WebumeniaMgImporter::class,
+            ]),
```
eronisko, line 36: I normally don't do 'random' in tests. It might make them unpredictable -- they can pass one time and then fail another time. If you want to test the behaviour for all the importers, I'd call the factory using each one of them.

However, in this instance I believe picking just one of them (e.g. the most important/common one) is a good enough approach.

database/migrations/2022_11_03_173406_cleanup_imports_table.php
```diff
@@ -0,0 +1,32 @@
...
+return new class extends Migration {
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('imports', function (Blueprint $table) {
+            $table->dropColumn(['settings', 'mapping']);
```
eronisko, line 16: Nice!

eronisko (approved):

PR #788, opened 2022-11-10
add temporary alert

igor-kamil:
# Description

Snímka obrazovky 2022-11-10 o 16 51 06


eronisko (approved):
As an alert it's fine. I just hope the outage doesn't generate some unexpected failures

PR #794, opened 2022-11-28
Webumenia 1916 include date earliest latest in items api

igor-kamil:
# Description

- include `date_earliest` and `date_latest` in ItemResource fro API v2
- fix mapping of item `dating` in harvester

Before:
Snímka obrazovky 2022-11-28 o 14 36 45

After:
Snímka obrazovky 2022-11-28 o 14 29 51

Proposed change keeps whole `dating` as it came from OAI-PMH
If there is multiple values, break them in multiple lines.
E.g.:
```
'created' => [
                '1760/1760',
                '18. storočie, polovica, 1760 (originál)',
                '1860/1860',
                '21. storočie, 1. polovica, 2019 (zväčšenina)',
            ],
```
result
```
18. storočie, polovica, 1760 (originál)
21. storočie, 1. polovica, 2019 (zväčšenina)
```

Fixes [WEBUMENIA-1916](https://jira.sng.sk/browse/WEBUMENIA-1916)

## Type of change

# Checklist:

### DIFF
--- a/app/Item.php
+++ b/app/Item.php
@@ -489,7 +489,7 @@ class Item extends Model implements IndexableModel, TranslatableContract
         return $value;
     }
 
-    public function getDatingFormated($single_line = false)
+    public function getDatingFormated($single_line = true)
     {
         $count_digits = preg_match_all("/[0-9]/", $this->dating);
         if (($count_digits<2) && !empty($this->date_earliest)) {
--- a/app/Item.php
+++ b/app/Item.php
@@ -489,7 +489,7 @@ class Item extends Model implements IndexableModel, TranslatableContract
         return $value;
     }
 
-    public function getDatingFormated()
+    public function getDatingFormated($single_line = false)
     {
         $count_digits = preg_match_all("/[0-9]/", $this->dating);
         if (($count_digits<2) && !empty($this->date_earliest)) {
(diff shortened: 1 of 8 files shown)


resources/views/dielo.blade.php
```diff
@@ -120,7 +120,7 @@ class="sr-only">{{ trans('dielo.item_next-work') }}</span></a>
                                 <td class="atribut">{{ trans('dielo.item_attr_dating') }}:</td>
                                 <td>
                                     <time itemprop="dateCreated" datetime="{{ $item->date_earliest }}">
-                                        {{ $item->getDatingFormated() }}
+                                        {!!nl2br($item->getDatingFormated()) !!}
```
rastislav-chynoransky, line 123: This is not the only occurrence of `Item::getDatingFormated`. How should display the rest?

igor-kamil: good point! fixed ✔️

app/Item.php
```diff
@@ -489,7 +489,7 @@ private function getMeasurementForDimension($dimension)
         return $value;
     }
 
-    public function getDatingFormated()
+    public function getDatingFormated($single_line = false)
```
rastislav-chynoransky, line 492: I would consider making the default `true`. There are currently 10 occurences of this method call in the codebase and all those expect single line.

rastislav-chynoransky: And there are also `$item->dating` calls. I think this needs a bit of refactoring...
If we need a quick fix, I'd save it to DB imploded with `;` and then format it only in those few cases.

igor-kamil: you are right... applied

chynoransky: Thanks!

rastislav-chynoransky: Thanks!

rastislav-chynoransky (approved):

PR #798, opened 2022-12-14
[import] fix prefixing of iipimg_url

rastislav-chynoransky:
We need to store in `item_images.iipimg_url` only unprefixed paths to JP2 files.


igor-kamil (approved):
LGTM

PR #803, opened 2023-01-23
Parse collection id inside the article

FrantisekMichalSebestyen:
# Description

## Type of change

# Checklist:

Snímka obrazovky 2023-01-23 o 21 11 22
Snímka obrazovky 2023-01-23 o 21 11 53


app/Article.php
```diff
@@ -146,16 +146,42 @@ public function getReadingTimeAttribute()
...
+
+                $collection_teaser = Blade::render(
+                    '<x-collection_teaser :collection="$collection" />',
+                    [
+                        'collection' => $collection,
+                    ]
+                );
+                return $collection_teaser;
+            });
         return $content;
```
eronisko, line 185: I would really like all this parsing and rendering logic to be tucked in a dedicated component. 

Something like this:
```suggestion
        return Str::of(html_entity_decode($this->content))
            ->replaceMatches('/\[article_teaser id=(.*?)\]/', function ($match) {
               return Teaser::buildFromMarkup($match[0])::render()
            });
```

Then this component would have things like
```php
namespace App\View\Components;

use Illuminate\View\Component;

class Teaser extends Component
{
    private string $type;
    private int $id;

    /**
     * Create a new component instance.
     *
     * @return void
     */
    public function __construct() {
        // ...
    }

    public static function buildFromMarkup(string $markup): Teaser
    {
       //... all the parsing happens here
    }

    public function render()
    {
        // rendering and probably even data fetching happens here
        if ($this->type === 'article') {
          // render x-article_teaser ...
        }
        // render something else
        // or render empty string
    }
}
```

resources/views/components/collection_teaser.blade.php
```diff
@@ -0,0 +1,43 @@
+@props(['collection'])
```
eronisko, line 1: Separate component here is fine, IMHO 👍

eronisko (changes requested):
I'm usually one for copy-pasting, but here I think we should merge the new CKEditor plugin into the previous (article) plugin.

I'm also suggesting moving the parsing logic into a dedicated component -- I think it could work nicely.

app/Article.php
```diff
@@ -146,16 +146,11 @@ public function getReadingTimeAttribute()
...
-            
-            $article_teaser = Blade::render('<x-article_teaser :article="$article" />', ['article' => $article]);
-            return $article_teaser;
-        });
-        return $content;
+        return Str::of(html_entity_decode($this->content))->replaceMatches(
+            '/\[teaser id=(.*?) type=(.*?)\]/',
+            function ($match) {
+                return Teaser::buildFromMarkup($match)->render();
+            }
```
eronisko, line 153:
```suggestion
            '/\[teaser.*?\]/',
            function ($match[0]) {
                return Teaser::buildFromMarkup($match[0])->render();
            }
```

app/View/Components/Teaser.php
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    public static function buildFromMarkup($match): Teaser
+    {
+        $id = (int) $match[1];
+        $type = Str::of($match[2]);
+
+        if (empty($id) || empty($type)) {
+            return '';
+        }
```
eronisko, line 34:
```suggestion
    {
        // ... match regex '/\[teaser id=(\d+) type=(article|collection)\]/'
        // if does not match, return ''
        $id = (int) $match[1];
        $type = Str::of($match[2]);

        if (empty($id) || empty($type)) {
            return '';
        }
```

app/View/Components/Teaser.php
```diff
@@ -0,0 +1,55 @@
...
+                'article' => $article,
+            ]);
+        } elseif ($this->type === 'collection') {
+            $collection = Collection::published()->find($this->id);
+            return Blade::render('<x-collection_teaser :collection="$collection" />', [
+                'collection' => $collection,
+            ]);
+        } else {
+            return '';
+        }
```
eronisko, line 53:
```suggestion
        if ($this->type === 'article') {
            $article = Article::published()->find($this->id);
            return Blade::render('<x-article_teaser :article="$article" />', [
                'article' => $article,
            ]);
        }
        if ($this->type === 'collection') {
            $collection = Collection::published()->find($this->id);
            return Blade::render('<x-collection_teaser :collection="$collection" />', [
                'collection' => $collection,
            ]);
        }
        return '';
```

app/View/Components/Teaser.php
```diff
@@ -0,0 +1,55 @@
...
+
+        return new Teaser($id, $type);
+    }
+
+    public function render()
+    {
+        if ($this->type === 'article') {
+            $article = Article::published()->find($this->id);
```
eronisko, line 42: Account for `$article` (`$collection`) not found

eronisko (approved):

PR #808, opened 2023-02-10
Feature/pnp 68 webdav flysystem

rastislav-chynoransky:
We've encountered an issue of davfs service serving cached version of files.
So far I haven't found a way to refresh its contents.
Therefore here we connect directly to webdav endpoint from the app.


app/Providers/AppServiceProvider.php
```diff
@@ -65,6 +70,20 @@ public function boot()
...
+                'userName' => $config['username'],
+                'password' => $config['password'],
+            ]);
+            $adapter = new WebDAVAdapter($client, $config['prefix'] ?? '');
+            $driver = new Filesystem($adapter);
+            return new FilesystemAdapter($driver, $adapter, $config);
+        });
+        if (app()->environment(['local', 'testing'])) {
+            Storage::fake('import_iip');
+        }
```
eronisko, line 86: This is a bit of a 'gotcha' to me. I wouldn't expect to see `fake()` used outside of tests.

I think a more standard way is to use a different disk by default by taking disk name here
https://github.com/SlovakNationalGallery/webumenia.sk/blob/3735c2e31db6d06db728ab92468a85056562de75/app/ImportRecord.php#L47 from config. 

That config would then be overridden in production and `fake`d in tests. Would that make sense to you?

eronisko (approved):
Nice, thanks. I'll release it right away

PR #825, opened 2023-03-16
Webumenia 1904 - Pagination + a bit of mansonry

FrantisekMichalSebestyen:
# Description

## Type of change

# Checklist:


resources/js/app.js
```diff
@@ -49,6 +48,7 @@ Vue.component('filter-disclosure-view', require('./components/catalog-new/Disclo
 Vue.component('filter-disclosure-controller', require('./components/catalog-new/DisclosureModalController.vue').default);
 Vue.component('filter-new-popover-group-controller', require('./components/catalog-new/PopoverGroupController.vue').default);
 Vue.component('filter-show-more', require('./components/catalog-new/ShowMore.vue').default);
+Vue.component('filter-masonry', require('./components/catalog-new/ArtworksMasonry.vue').default);
```
eronisko, line 51:
```suggestion
Vue.component('items-masonry', require('./components/catalog-new/ItemsMasonry.vue').default);
```

resources/js/components/catalog-new/ArtworkImage.vue
```diff
@@ -0,0 +1,34 @@
+<template>
+    <div>
+        <img :class="{ 'tw-hidden': !this.isLoaded }" @load="onImgLoad" :src="`${src}/600`" />
+        <div
+            :class="[{ 'tw-hidden': this.isLoaded }, 'tw-w-full tw-saturate-50']"
+            :style="{
+                'aspect-ratio': ratio || 1,
+                'background-color': hsl.length
+                    ? `hsl(${hsl[0].h}, ${hsl[0].s}%, ${hsl[0].l}%)`
```
eronisko, line 9: Please pass just one color as the prop so we don't have to look up `[0]` everywhere

resources/js/components/catalog-new/ArtworkImage.vue
```diff
@@ -0,0 +1,34 @@
...
+                    : '#cdcdcd',
```
eronisko, line 10: This could be set in a css class instead

resources/js/components/catalog-new/ArtworkImage.vue
```diff
@@ -0,0 +1,34 @@
...
+            }"
+        ></div>
+    </div>
+</template>
+
+<script>
+export default {
+    props: {
+        src: String,
+        ratio: Number,
```
eronisko, line 20:
```suggestion
        aspectRatio: Number,
```

resources/js/components/catalog-new/ArtworkImage.vue
```diff
@@ -0,0 +1,34 @@
...
+                    : '#cdcdcd',
+            }"
+        ></div>
+    </div>
+</template>
+
+<script>
+export default {
+    props: {
+        src: String,
```
eronisko, line 19: It would be also good to add `srcSet` here

resources/js/components/catalog-new/ArtworkImage.vue
```diff
@@ -0,0 +1,34 @@
...
+        ratio: Number,
+        hsl: Array,
```
eronisko, line 21:
```suggestion
        placeholderColorHsl: Object 
```

resources/js/components/catalog-new/ArtworksMasonry.vue
```diff
@@ -0,0 +1,55 @@
...
+                <ArtworkImage
+                    :src="`${url}${artwork.id}`"
+                    :ratio="artwork.content.image_ratio"
+                    :hsl="artwork.content.hsl"
+                ></ArtworkImage>
+                <span>{{ artwork.content.title }}</span>
+            </div>
+        </div>
+        <div>
+            <button :class="{ 'tw-invisible': this.page }" ref="more" @click="$emit('loadmore')">
```
eronisko, line 19:
```suggestion
            <button v-if="this.page" ref="loadMoreButton" @click="$emit('loadmore')">
```

resources/js/components/catalog-new/ArtworksMasonry.vue
```diff
@@ -0,0 +1,55 @@
...
+<script>
+import { VueMasonryPlugin } from 'vue-masonry'
+import ArtworkImage from './ArtworkImage.vue'
+
+Vue.use(VueMasonryPlugin)
+
+export default {
+    props: {
+        artworks: Array,
+        url: String,
```
eronisko, line 35: Please pick a more specific name, it's not clear what `url` this is.

resources/js/components/catalog-new/ArtworksMasonry.vue
eronisko, line 19: IMHO this button does not need to be a part of this component. It could be a separate "loadMore" component that both renders the button and does the auto-load.

And then we can actually take everything else from ArtworksMasonry out and into the Blade template.

resources/js/components/catalog-new/ItemsFilterController.vue
```diff
@@ -161,40 +163,49 @@ export default {
...
                             },
                         })
                     )
                     .then(({ data }) => data)
 
-                this.filters = {
-                    ...this.filters,
-                    ...filters,
+                if (replaceArtworks) {
+                    this.page = null
```
eronisko, line 186: When I see a side-effect like this, that's usually a sign that something is not quite right.

I would expect something like:
```js
const artworks = await axios.get(...)
const filters = await axios.get(...)

if (append) {
  this.artworks = [...this.artworks, ...artworks]
  this.filters = [...this.filters, ...filters]
  return
}

this.artworks = artworks
this.filters = filters
```

resources/js/components/catalog-new/ItemsFilterController.vue
```diff
@@ -203,8 +214,13 @@ export default {
         },
     },
     watch: {
+        page(newPage, oldPage) {
+            if (newPage > oldPage) {
+                this.fetchData({ replaceArtworks: false })
+            }
+        },
```
eronisko, line 221: Wouldn't the data get fetched twice? This doesn't seem like the right pattern here. This kind of logic should probably happen in `incrementPage` or in the `query` observer

resources/views/frontend/catalog-new/index-new.blade.php
```diff
@@ -204,51 +212,57 @@ class="tw-font-bold">(@{{ artworks.total }})</span>
...
+                            podľa</span>
+                        <span v-else>Zobrazujem <span
+                                class="tw-font-bold">@{{ artworks['total'] }}</span>
+                            diel, zoradených
+                            podľa</span>
+                    </filter-new-sort>
+                    <div class="tw-min-h-screen">
+                        <filter-masonry v-if="!isFetching || page" :key="JSON.stringify(query)"
+                            url="https://www.webumenia.sk/dielo/nahlad/" :page="page"
+                            @loadmore="incrementPage" :artworks="artworks"></filter-masonry>
```
eronisko, line 264:
```suggestion
                            @loadMore="loadMore" :artworks="artworks"></filter-masonry>
```

resources/js/components/catalog-new/ItemsFilterController.vue
```diff
@@ -220,10 +236,13 @@ export default {
     render() {
         return this.$scopedSlots.default({
             isExtendedOpen: this.isExtendedOpen,
+            isFetching: this.isFetching,
             query: this.query,
+            page: this.page,
```
eronisko, line 241: I think `page` should be a part of the `query`

eronisko (changes requested):
Nice to see this work! I just have some code organization/naming notes.

resources/views/frontend/catalog-new/index-new.blade.php
```diff
@@ -204,51 +212,57 @@ class="tw-font-bold">(@{{ artworks.total }})</span>
...
+                                class="tw-font-bold">@{{ artworks['total'] }}</span> diela, zoradené
+                            podľa</span>
+                        <span v-else>Zobrazujem <span
+                                class="tw-font-bold">@{{ artworks['total'] }}</span>
+                            diel, zoradených
+                            podľa</span>
+                    </filter-new-sort>
+                    <div class="tw-min-h-screen">
+                        <filter-masonry v-if="!isFetching || page" :key="JSON.stringify(query)"
+                            url="https://www.webumenia.sk/dielo/nahlad/" :page="page"
```
eronisko, line 263: Please use the Laravel `route` helper here

resources/js/components/catalog-new/ArtworksMasonry.vue
FrantisekMichalSebestyen: Not gonna work, the element is needed for IntersectionObserver:
However I can do something like this, If you prefer:
```suggestion
            <div ref="loadMoreWrapper"> 
                <button v-if="this.page" ref="loadMoreButton" @click="$emit('loadmore')">
           </div>
```

eronisko: Ah, gotcha. Would it work with `v-show`? If not, then let's out the observer on an element that's not designed to hide, like something with zero-height.

resources/js/components/catalog-new/ItemsFilterController.vue
```diff
@@ -161,40 +163,49 @@ export default {
                 [name]: this.query[name].filter((v) => v !== value),
             }
         },
-        async fetchData() {
+        incrementPage() {
+            this.page = this.page ? this.page + 1 : 2
+        },
+        async fetchData({ replaceArtworks }) {
```
eronisko, line 169: Maybe separate into two different methods
```
fetchArtworks(append = false)
fetchAggregations()
```

resources/js/components/catalog-new/ArtworksMasonry.vue
FrantisekMichalSebestyen: not working with `v-show` will use the second option

resources/js/components/catalog-new/ArtworkImage.vue
FrantisekMichalSebestyen: I would prefer to do it in separate PR

FrantisekMichalSebestyen: the one where i style it and add routing, favs, etc.

resources/js/app.js
FrantisekMichalSebestyen: got rid of it completely: Now its just handler of infinite scroll

eronisko (changes requested):
Some good progress, thanks. I have mostly minor suggestions

resources/js/components/catalog-new/ArtworkImage.vue
```diff
@@ -0,0 +1,43 @@
+<template>
+    <div>
+        <img :class="{ 'tw-hidden': !this.isLoaded }" @load="onImgLoad" :src="imageSrc" />
```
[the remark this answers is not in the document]
FrantisekMichalSebestyen: Is done in #826

resources/js/components/catalog-new/ItemsFilterController.vue
```diff
@@ -61,14 +62,16 @@ export default {
     data() {
         return {
             isExtendedOpen: true,
-            isFetching: false,
+            isFetchingArtworks: false,
```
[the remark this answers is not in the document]
FrantisekMichalSebestyen: Hm... I do not think it is used but, it might be needed for someone using the component, 🤷 Tbh, we should also do error handling. (we can talk about it). So you think i should get rid of is (fetchingArtworks completely) Or add loading insided the filters with aggregations ?

resources/js/components/catalog-new/ArtworkImage.vue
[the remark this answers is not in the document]
eronisko: And you did mention that previously, sorry

resources/js/components/catalog-new/ItemsFilterController.vue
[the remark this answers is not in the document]
eronisko: The ItemsFilterController is a very specific (non-reusable) component so it'll most likely only be used where it is used currently.

The point I was trying to make is that unless we actually need that variable, I wouldn't even create it. If we need it in the future, it's easy to add. But if we end up not using it, it's not easy to remember to delete it, and we're effectively creating technical debt.

Error handling -- I agree, but that would have to be done in a separate task since it's not designed right now.

eronisko (approved):
Looking good, thanks

PR #826, opened 2023-03-20
Masonry tile

FrantisekMichalSebestyen:
# Description

## Type of change

# Checklist:


FrantisekMichalSebestyen:
@eronisko Sorry, I cannot test it properly, i do not have image-previews locally. We tried to fix it with igor, but it did not work. Could you help me?

resources/views/frontend/catalog-new/index-new.blade.php
```diff
@@ -260,14 +260,40 @@ class="tw-font-bold">@{{ artworks['total'] }}</span>
...
+                                                </span>
+                                                <span v-else>
+                                                    <svg xmlns="http://www.w3.org/2000/svg" class="tw-w-5 tw-h-5 tw-fill-current" viewBox="0 0 256 256"><path d="M243,96.05a20,20,0,0,0-17.26-13.72l-57-4.93-22.3-53.14h0a20,20,0,0,0-36.82,0L87.29,77.4l-57,4.93A20,20,0,0,0,18.87,117.4l43.32,37.8-13,56.24A20,20,0,0,0,79,233.1l49-29.76,49,29.76a20,20,0,0,0,29.8-21.66l-13-56.24,43.32-37.8A20,20,0,0,0,243,96.05Zm-66.75,42.62a20,20,0,0,0-6.35,19.63l11.39,49.32-42.94-26.08a19.9,19.9,0,0,0-20.7,0L74.71,207.62,86.1,158.3a20,20,0,0,0-6.35-19.63L41.66,105.44,91.8,101.1a19.92,19.92,0,0,0,16.69-12.19L128,42.42l19.51,46.49A19.92,19.92,0,0,0,164.2,101.1l50.14,4.34Z"></path></svg>
+                                                </span>
+                                            </template>
+                                        </user-collections-favourite-button>
```
eronisko, line 294: Making the icons pluggable (into slots) is not the right approach in my opinion. It adds unnecessary complexity.

The component itself is so simple I'd just make a `FavouriteButtonV2` copy component with the right icons. We could DRY up the business logic there, but I don't think it's worth the effort in this case.

resources/views/frontend/catalog-new/index-new.blade.php
```diff
@@ -260,14 +260,40 @@ class="tw-font-bold">@{{ artworks['total'] }}</span>
...
                                 :key="artwork.id">
-                                <filter-artwork-image name="artwork-image" :artwork-id="artwork.id"
-                                    preview-url-template="{{ route('dielo.nahlad', ['id' => '__ARTWORK_ID__', 'width' => '__WIDTH__']) }}"
+                                <filter-artwork-tile name="artwork-image" :artwork-id="artwork.id"
+                                    preview-url-template="{{ route('dielo.nahlad', ['id' => '__ARTWORK_ID__', 'width' => '600']) }}"
+                                    preview-src-set-template="{!! route('dielo.nahlad', ['id' => '__ARTWORK_ID__', 'width'=> '600']) !!} 600w,
+                                        {!! route('dielo.nahlad', ['id' => '__ARTWORK_ID__', 'width'=>'220']) !!} 220w,
+                                        {!! route('dielo.nahlad', ['id' => '__ARTWORK_ID__', 'width'=>'300']) !!} 300w,
+                                        {!! route('dielo.nahlad', ['id' => '__ARTWORK_ID__', 'width'=>'600']) !!} 600w,
+                                        {!! route('dielo.nahlad', ['id' => '__ARTWORK_ID__', 'width'=>'800']) !!} 800w"
```
eronisko, line 271: Are the `{!! !!}` really necessary here?
Also, the image width (e.g. `220`) is probably not the same as screen width (e.g. `220w`) if we're showing the images in columns.

resources/js/components/catalog-new/ArtworkTile.vue
```diff
@@ -0,0 +1,99 @@
...
+                :srcSet="imageSrcSet"
+            />
+        </a>
+        <div
+            :class="[{ 'tw-hidden': this.isLoaded }, 'tw-w-full tw-saturate-50']"
+            :style="{
+                'aspect-ratio': aspectRatio || 1,
+                'background-color': placeholderColorHsl
+                    ? `hsl(${placeholderColorHsl.h}, ${placeholderColorHsl.s}%, ${placeholderColorHsl.l}%)`
+                    : '#cdcdcd',
```
eronisko, line 17: CSS class? :)

resources/js/components/catalog-new/ArtworkTile.vue
```diff
@@ -0,0 +1,99 @@
...
+                    >{{ title }}</a
+                >
+                <a
+                    :href="artworkUrl"
+                    class="tw-block tw-leading-5 tw-pb-2 tw-font-normal tw-text-base"
+                    >{{ dating }}</a
+                >
+            </div>
+            <div class="tw-flex tw-gap-4 tw-items-start">
+                <slot name="favourite-button"></slot>
```
eronisko, line 39: This is my beef with Vue components... It's silly to pass the favourite-button through a slot, there's no need for the flexibility -- it should always look/be the same. But we can't just put it in because we don't have translation strings available in Vue. So we either do a slot or we do some prop drilling.

Looking at the props of this component, it really is just concerned with parsing things from Laravel into Vue so I'd suggest skipping that part altogether.

I think it'll be better to make this into either a Blade component or no component (i.e. inline it). I'd keep the `<img` and the placeholder in a Vue component (to keep the interactive feature), and everything else can stay in Blade.

resources/js/components/catalog-new/ArtworkImageController.vue
FrantisekMichalSebestyen, line 1: Not sure where to put this. As this will be probably used all over webumenia.

resources/js/components/catalog-new/FavouriteController.vue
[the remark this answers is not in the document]
eronisko: We could go one level deeper and instead of exposing `store`, expose `toggleItem`, `hasItem`, etc

eronisko (changes requested):
It looks good overall, but I suspect it has some changes carried over from #825.

Please merge/re-base and I'll review again.

resources/js/components/catalog-new/FavouriteController.vue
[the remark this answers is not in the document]
eronisko: Also then it could be called user-collections-store (and be stored as such, under `components/user-collections`)

resources/views/frontend/catalog-new/index-new.blade.php
```diff
@@ -204,51 +212,139 @@ class="tw-font-bold">(@{{ artworks.total }})</span>
...
+                                            <a :href="route('dielo', {id: artwork.id})"
+                                                class="tw-pb-2 tw-text-lg tw-font-medium tw-leading-5">@{{ artwork.content.title }}</a>
+                                            <a :href="route('dielo', {id: artwork.id})"
+                                                class="tw-pb-2 tw-text-base tw-font-normal tw-leading-5">@{{ artwork.content.dating }}</a>
+                                        </div>
+                                        <div class="tw-flex tw-items-start tw-gap-4">
+                                            <favourite-controller v-slot="{ store }">
+                                                <button @click="store.toggleItem(artwork.id)">
+                                                    <span v-if="store.hasItem(artwork.id)">
+                                                        <svg xmlns="http://www.w3.org/2000/svg"
```
eronisko, line 302: probably no need for the spans here -- just `v-if` the svg
```suggestion
                                                    <svg v-if="store.hasItem(artwork.id)" xmlns="http://www.w3.org/2000/svg"
```

PR #830, opened 2023-03-31
Update PHPUnit to v10

eronisko:
## Type of change

# Checklist:

### DIFF
--- a/tests/Harvest/Importers/AuthorityImporterTest.php
+++ b/tests/Harvest/Importers/AuthorityImporterTest.php
@@ -14,30 +14,29 @@ use App\Harvest\Mappers\NationalityMapper;
 use App\Harvest\Mappers\RelatedAuthorityMapper;
 use App\Harvest\Progress;
 use App\Nationality;
-use Elasticsearch\Client;
 use Illuminate\Foundation\Testing\RefreshDatabase;
+use Tests\Harvest\FakeRecordFactory;
 use Tests\TestCase;
 
 class AuthorityImporterTest extends TestCase
 {
+    private AuthorityImporter $importer;
+
     use RefreshDatabase;
 
     protected function setUp(): void
     {
         parent::setUp();
-        $this->app->instance(Client::class, $this->createMock(Client::class));
-    }
 
-    public function testUpdateRelated()
-    {
-        Authority::factory()->create(['id' => 1000162]);
-        Authority::factory()->create(['id' => 1000168]);
-        Authority::factory()->create(['id' => 11680]);
-
-        $row = $this->getData();
-        $importer = $this->initImporter($row);
-
-        $authority = $importer->import($row, new Progress());
+        $this->importer = new AuthorityImporter(
+            app(AuthorityMapper::class),
+            app(AuthorityEventMapper::class),
+            app(AuthorityNameMapper::class),
+            app(AuthorityNationalityMapper::class),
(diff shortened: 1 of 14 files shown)


tests/Harvest/Importers/AuthorityImporterTest.php
```diff
@@ -119,115 +81,9 @@ protected function getData()
...
             ],
-            'links' => [
-                [
-                    'url' => ['http://example.org/'],
-                ],
-            ],
-        ];
-    }
-
-    protected function initImporter(array $row)
```
eronisko, line 130: I found the the mocking here unnecessary. The mappers don't have any dependencies so I've decided to just initialize them normally.

I'm also mocking the importer data from the new `FakeRecordFactory`. @rastislav-chynoransky, if you can suggest a better name for that factory, I'd be happy to adjust it.

tests/Observers/AuthorityObserverTest.php
```diff
@@ -130,32 +130,8 @@ public function testUpdateRelatedWhenSynced()
...
-        $authorityElasticsearchMock
-            ->expects($this->at(1))
-            ->method('index')
-            ->with(
-                $this->callback(function (array $params) {
-                    return 1 === $params['body']['items_count'];
-                })
-            );
+        $itemElasticsearchMock->expects($this->atLeast(2))->method('index');
+        $authorityElasticsearchMock->expects($this->atLeastOnce())->method('index');
```
eronisko, line 134: The [`at`](https://www.drupal.org/node/3218874) and [`withConsecutive`](https://github.com/sebastianbergmann/phpunit/issues/4026) methods are removed from PHPUnit 10.

As of now, [there is no 1:1 replacement](https://www.drupal.org/project/drupal/issues/3306554) for these, so I've decided to simplify these tests and leave out the checking of the parameters of the `index` calls.

I believe, however, that these tests are sufficient -- we want to know whether indexing is triggered, and the validity of the indexing itself is covered by other tests.

tests/Harvest/Importers/AuthorityImporterTest.php
```diff
@@ -14,30 +14,29 @@
...
 class AuthorityImporterTest extends TestCase
 {
+    private AuthorityImporter $importer;
+
     use RefreshDatabase;
 
     protected function setUp(): void
     {
         parent::setUp();
-        $this->app->instance(Client::class, $this->createMock(Client::class));
```
rastislav-chynoransky, line 28: Will it now index the data? If so could we reduce the unnecessary work of tests?

tests/Harvest/Importers/AuthorityImporterTest.php
rastislav-chynoransky: It could be split into `ItemRecordFactory` and `AuthorityRecordFactory`, but I'm also okay with your proposal.

rastislav-chynoransky (approved):
Thanks! 🧹 ✨

tests/Harvest/Importers/AuthorityImporterTest.php
eronisko: Good point! Putting this back didn't work for some reason. I created the `WithoutSearchIndexing` trait in [6700ea7](https://github.com/SlovakNationalGallery/webumenia.sk/pull/830/commits/6700ea7cdc7c5ab2fa4cfb6ea87fb7e2417da006) that does disable the observers.

PR #856, opened 2023-05-12
Add authors_formatted field to /api/v1/items

eronisko:
This extends the V1 API instead of creating a V2, as creating a V2 would be too big a commitment :)

## Type of change

# Checklist:

### DIFF
--- a/app/Http/Controllers/Api/V1/ItemController.php
+++ b/app/Http/Controllers/Api/V1/ItemController.php
@@ -14,7 +14,6 @@ use Primal\Color\Parser as ColorParser;
 
 class ItemController extends Controller
 {
-
     private $filterables = [
         'author',
         'topic',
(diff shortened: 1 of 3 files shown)


app/Http/Controllers/Api/V1/ItemController.php
```diff
@@ -75,7 +70,25 @@ public function index(Request $request)
                 $searchRequest->sort($field, $direction);
             });
 
-        return $searchRequest->paginate($size)->onlyDocuments();
+        $response = $searchRequest
+            ->paginate($size)
+            ->onlyDocuments()
+            ->toArray();
+
+        return [
```
FrantisekMichalSebestyen, line 78: This looks really ugly, but I will trust you there is no better way to code this :)

FrantisekMichalSebestyen (approved):

app/Http/Controllers/Api/V1/ItemController.php
eronisko: I tried but I can't come up with anything less obnoxious right now...

PR #858, opened 2023-05-12
feat: add loading indication on images of artworks

FrantisekMichalSebestyen:
Approved by Gorazd

https://github.com/SlovakNationalGallery/webumenia.sk/assets/28804184/e5d8892a-cf06-43b7-9a8e-e3000e5d52e9


eronisko (approved):
I still think that plain white would've been better, but code looks good :)

PR #861, opened 2023-05-17
feat: hide filters on empty aggregation results

FrantisekMichalSebestyen:
# Description
Desktop:
https://github.com/SlovakNationalGallery/webumenia.sk/assets/28804184/f75f652d-0446-482e-86af-c9e84d1e132b
Mobile:
https://github.com/SlovakNationalGallery/webumenia.sk/assets/28804184/a112d83a-f151-4fa2-8064-6a346450979a

### DIFF
--- a/resources/js/components/catalog-new/ItemsFilterController.vue
+++ b/resources/js/components/catalog-new/ItemsFilterController.vue
@@ -162,6 +162,12 @@ export default {
         },
     },
     methods: {
+        isFilterEmpty(filterName) {
+            return (
+                this.query[filterName].length ||
+                (this.aggregations[filterName] && this.aggregations[filterName].length > 0)
+            )
+        },
         clearFilterSelection(filterName) {
             this.query = {
                 ...this.query,
@@ -321,6 +327,7 @@ export default {
             aggregations: this.aggregations,
             artworks: this.artworks,
             artworks_total: this.artworks_total,
+            isFilterEmpty: this.isFilterEmpty,
             loadMore: this.loadMore,
             handleSortChange: this.handleSortChange,
             handleYearRangeChange: this.handleYearRangeChange,
--- a/resources/js/components/catalog-new/ItemsFilterController.vue
+++ b/resources/js/components/catalog-new/ItemsFilterController.vue
@@ -162,7 +162,7 @@ export default {
         },
     },
     methods: {
-        isFilterEmpty(filterName) {
+        hasFilterOptions(filterName) {
             return (
                 this.query[filterName].length ||
                 (this.aggregations[filterName] && this.aggregations[filterName].length > 0)
@@ -327,7 +327,7 @@ export default {
             aggregations: this.aggregations,
             artworks: this.artworks,
             artworks_total: this.artworks_total,
-            isFilterEmpty: this.isFilterEmpty,
+            hasFilterOptions: this.hasFilterOptions,
             loadMore: this.loadMore,
             handleSortChange: this.handleSortChange,
             handleYearRangeChange: this.handleYearRangeChange,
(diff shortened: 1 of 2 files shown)


eronisko (changes requested):
Hm, this is probably too repetitive even to my liking... You know I'm not a DRY freak, but here we're repeating the same logic over and over again. Let's think about this one some more.

resources/js/components/catalog-new/NewColorSlider.vue
```diff
@@ -41,7 +43,7 @@
             :max="1"
             :interval="0.01"
             :duration="0"
-            :dotSize="24"
+            :dotSize="44"
             :height="12"
```
eronisko, line 47: These changes probably don't belong here, please double check your commit history

resources/js/components/catalog-new/NewCustomSelectPopoverLabel.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-    <span class="tw-font-semibold tw-text-sm md:tw-text-base tw-flex tw-gap-1">
+    <span class="tw-font-semibold tw-text-base tw-flex tw-gap-1">
```
eronisko, line 2:
```suggestion
    <span class="tw-font-semibold tw-flex tw-gap-1">
```

resources/js/components/catalog-new/ItemsFilterController.vue
```diff
@@ -162,6 +162,12 @@ export default {
         },
     },
     methods: {
+        isFilterEmpty(filterName) {
+            return (
+                this.query[filterName].length ||
+                (this.aggregations[filterName] && this.aggregations[filterName].length > 0)
+            )
+        },
```
eronisko, line 170: Should be under `computed`.

Also the name doesn't seem to fit what it does. How about `filterHasOptions` ?

eronisko (changes requested):
Looks like there are lot of unrelated changes here, please check

resources/js/components/catalog-new/ItemsFilterController.vue
FrantisekMichalSebestyen: Computed does not accept parameters:
I can return function: filterHasOptions{return (param) => ...}
However then there are no benefits to caching and we should use methods
More importanly: **Do not use computed properties to mutate DOM**
So in worst case we can break reactivity in vue
Snímka obrazovky 2023-05-31 o 13 03 42
Src: https://vuejs.org/guide/essentials/computed.html#writable-computed

eronisko: You're right, I forgot that `computed` didn't accept params. (The other aspects are not a concern -- we're not producing any side effects)

resources/views/frontend/catalog-new/index-new.blade.php
```diff
@@ -129,7 +133,8 @@ class="tw-mt-4 tw-flex tw-h-[30rem] tw-w-[20rem] tw-flex-col tw-items-start tw-b
                                         @endslot
                                     </x-filter.search_popover>
                                     <x-filter.search_popover v-bind:is-active="query.topic.length > 0"
-                                        v-if="tc.isOn" name="topic">
+                                        v-if="(hasFilterOptions('topic') && tc.isOpen)" v-if="tc.isOn"
```
eronisko, line 136: `v-if` is here (and in other places) twice -- please check

eronisko (changes requested):
Looking much better, thanks. Please check for redundant `v-if` on the modified components

resources/views/frontend/catalog-new/index-new.blade.php
FrantisekMichalSebestyen: uff ur right 🤦

eronisko (approved):
Nice, thanks

eronisko (approved):
Nice, thanks

eronisko (approved):
Nice, thanks

PR #865, opened 2023-05-31
error message + no option found dep. on search field

FrantisekMichalSebestyen:
Snímka obrazovky 2023-05-31 o 15 17 07

Snímka obrazovky 2023-05-31 o 15 22 34

### DIFF
--- a/resources/js/components/catalog-new/ItemsFilterController.vue
+++ b/resources/js/components/catalog-new/ItemsFilterController.vue
@@ -112,6 +112,7 @@ export default {
     data() {
         return {
             isFetchingArtworks: false,
+            hasError: false,
             artworks: [],
             last_page: 1,
             artworks_total: null,
@@ -264,7 +265,7 @@ export default {
                     ...aggregations,
                 }
             } catch (e) {
-                throw e
+                this.hasError = true
             }
         },
         async fetchArtworks({ append }) {
@@ -290,7 +291,7 @@ export default {
                     ? [...this.artworks, ...fetchedArtworks.data]
                     : fetchedArtworks.data
             } catch (e) {
-                throw e
+                this.hasError = true
             } finally {
                 this.isFetchingArtworks = false
             }
@@ -321,6 +322,7 @@ export default {
     render() {
         return this.$scopedSlots.default({
             isFetchingArtworks: this.isFetchingArtworks,
+            hasError: this.hasError,
             query: this.query,
             page: this.page,
             last_page: this.last_page,
--- a/resources/js/components/catalog-new/ItemsFilterController.vue
+++ b/resources/js/components/catalog-new/ItemsFilterController.vue
@@ -266,6 +266,7 @@ export default {
                 }
             } catch (e) {
                 this.hasError = true
+                throw e
             }
         },
         async fetchArtworks({ append }) {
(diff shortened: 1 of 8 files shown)


resources/views/frontend/catalog-new/index-new.blade.php
```diff
@@ -696,7 +697,15 @@ class="tw-h-4 tw-w-4 tw-fill-current" viewBox="0 0 256 256">
...
                     <div class="tw-min-h-screen">
                         <div class="tw-px-4 md:tw-px-16 md:tw-py-10">
-                            <div v-if="artworks.length === 0"
+                            <div v-if="hasError"
+                                class="tw-flex tw-w-full tw-flex-col tw-items-center tw-justify-center tw-py-40 tw-text-lg">
+                                <span>Uuups, niečo sa pokazilo :( skús</span>
+                                <reload-controller v-slot="rc">
+                                    <button class="tw-inline-block tw-underline" @click="rc.reload">obnoviť
+                                        stránku.</button>
+                                </reload-controller>
```
eronisko, line 706: This is fine and consistent, but it could've just been a simple `<reload-button />` component. There is nothing to "control" here
```suggestion
                                <reload-button class="tw-inline-block tw-underline">
                                  obnoviť stránku.
                                </reload-button>
```

resources/views/components/filter/search_options.blade.php
```diff
@@ -1,46 +1,37 @@
...
+                    class="tw-form-checkbox tw-m-0 tw-mr-3 tw-h-6 tw-w-6 tw-border-gray-300 focus:tw-outline-none focus:tw-ring-0 focus:tw-ring-offset-0"
+                    type="checkbox" :key="option . id" :id="option . id" :value="option . value"
+                    :checked="option . checked" @change="{{ $attributes->get('v-on:change') }}" />
+                <span class="tw-text-base tw-font-normal">@{{ option.value }} <span
+                        class="tw-font-semibold">(@{{ option.count }})</span>
                 </span>
             </label>
+            <div v-else class="tw-flex tw-flex-col tw-items-center tw-justify-center tw-gap-4">
+                <lottie-player autoplay loop mode="normal"
+                    src="{{ asset('animations/empty.json') }}">
```
eronisko, line 27: Didn't we show this animation somewhere else already?

resources/js/components/catalog-new/ItemsFilterController.vue
```diff
@@ -258,7 +259,7 @@ export default {
                     ...aggregations,
                 }
             } catch (e) {
-                throw e
+                this.hasError = true
```
eronisko, line 262: It would be good to at least log the error into console. Would this work?
```
this.hasError = true
throw e
```

eronisko (changes requested):
It looks good, just please fix the error handling so that we don't lose the error completely

resources/views/components/filter/search_options.blade.php
FrantisekMichalSebestyen: yup, should I extract it?

resources/js/components/catalog-new/ItemsFilterController.vue
```diff
@@ -258,6 +259,7 @@ export default {
                     ...aggregations,
                 }
             } catch (e) {
+                this.hasError = true
```
eronisko, line 262: Can this actually ever recover from `hasError: true`?

FrantisekMichalSebestyen: Not really, should it?

eronisko (approved):

PR #878, opened 2023-07-20
Fix add zoom to api

FrantisekMichalSebestyen:


tests/Feature/Api/V2/ItemsTest.php
```diff
@@ -53,6 +63,7 @@ public function test_detail()
                 'date_latest' => 2010,
                 'description' => 'test_description',
                 'image_ratio' => 1.5,
+                'images_zoom' => [$image_zoom],
```
eronisko, line 66: Could you please put the expected result here instead of computing it? (When code under test is duplicated in the test, we get into a weird tautology situation)

app/Http/Resources/Api/V2/ItemResource.php
```diff
@@ -14,6 +15,8 @@ class ItemResource extends JsonResource
      */
     public function toArray($request)
     {
+        $this->loadMissing('images');
```
eronisko, line 18: I see why this makes sense but this kind of code usually belongs in the controller. (The controller is responsible for querying/data fetching)

app/Http/Resources/Api/V2/ItemResource.php
```diff
@@ -24,6 +27,9 @@ public function toArray($request)
             'description' => $this->description,
             'authorities' => AuthorityResource::collection($this->authorities),
             'image_ratio' => $this->image_ratio,
+            'images_zoom' => $this->images->map(function (ItemImage $image) {
+                return $image->getDeepZoomUrl();
+            }),
```
eronisko, line 32: We'll likely have more than 'deepZoomUrl' for images. Could we do something like this instead?

```json
"images": [
  {
    "deep_zoom_url": "http://...."
  }
]
```

eronisko (changes requested):
Thanks for this 👍

tests/Feature/Api/V2/ItemsTest.php
FrantisekMichalSebestyen: Hmm.... not really if I understand correctly. 
You want the test to be independent of .env values. So we need to load them and parse the url.

eronisko (approved):

PR #892, opened 2023-08-25
fix: regex on format author

FrantisekMichalSebestyen:


resources/js/components/FormatterController.vue
```diff
@@ -0,0 +1,13 @@
+<script>
```
FrantisekMichalSebestyen, line 1: @eronisko I left here alternative proposal

eronisko: Thanks. When I think about formatters I usually try err on the side of not re-using enough, because to me, formatting is a design choice and can be different on different pages. So I try to keep my formatting utils as local as possible.

What I had in mind when I suggested a 'formatter component' was something like this:

```
<Catalog.AuthorName :value="author" />
<Catalog.ArtworksTotal :value="artworks_total" />
```

So not even a `<AuthorNameFormatter />` (which would suggest reusability) but rather a component specific to the page it is featured at.

Here's why: Imagine we have a page where we want to show author names using initials, say `R. Fila` in articles. Now `AuthorNameFormatter` needs to take a `format` parameter (or similar) in order to make it what it wants to be: a re-usable formatting component. That complicates the component's props as well as its internals. So the component becomes harder to use and harder to maintain. (This is a simple example but I hope I'm making my point)

On the other hand, having a `Catalog.AuthorName` and `Article.AuthorName` gives me two components with just one `value` prop,. Inside these components I can make a choice of re-using any formatting utils or not, but the user of the component does not have to care. If someone wants to re-use my `Catalog.AuthorName` in e.g. collections, they can either use it directly in their `Collection.AuthorName`, or make a more "global" `AuthorFullName` component.

resources/js/components/catalog-new/NumberFormatter.vue
```diff
@@ -0,0 +1,16 @@
+<template>
+    {{
+        value
+            .toLocaleString('en-US', {
+                useGrouping: true,
+            })
+            .replace(/,/g, ' ')
+    }}
```
eronisko, line 8: This is a bit awkward to read, could you please move it into `computed`?

I'd also change the locale so that we don't have to remove the comma:

```suggestion
    {{
        value
            .toLocaleString('sk-SK', {
                useGrouping: true,
            })
    }}
```

resources/views/components/filter/search_options.blade.php
```diff
@@ -1,26 +1,30 @@
...
-                    class="tw-inline-block tw-min-w-0 tw-break-words tw-text-base tw-font-normal">@{{ sc.formatter ? sc.formatter(option.value) : option.value }}
-                    <span class="tw-font-semibold">(@{{ option.count }})</span>
+                    type="checkbox" v-bind:key="option.id" v-bind:id="option.id"
+                    v-bind:value="option.value" v-bind:checked="option.checked"
+                    @change="{{ $attributes->get('v-on:change') }}" />
+                <span class="tw-inline-block tw-min-w-0 tw-break-words tw-text-base tw-font-normal">
+                    {!! $label ?? '@{{ option.value }}' !!}
+                    <span class="tw-font-semibold">
+                        (<catalog.number-formatter v-bind:value="option.count">
+                        </catalog.number-formatter>)
```
eronisko, line 26:
```suggestion
                        (<catalog.number-formatter v-bind:value="option.count" />)
```

resources/views/frontend/catalog-new/index-new.blade.php
```diff
@@ -479,10 +484,15 @@ class="tw-inset-x-0 tw-box-border tw-flex tw-min-h-0 tw-flex-1 tw-flex-col tw-ov
...
                                                             v-bind:options="aggregations['author']"
-                                                            v-bind:formatter="formatAuthor"
                                                             v-bind:selected="query['author']"
                                                             v-on:change="e => handleMultiSelectChange('author', e)"
-                                                            v-on:reset="clearFilterSelection('author')" />
+                                                            v-on:reset="clearFilterSelection('author')">
+                                                            @slot('label')
+                                                                <catalog.author-formatter
+                                                                    v-bind:value="option.value">
+                                                                </catalog.author-formatter>
```
eronisko, line 493:
```suggestion
                                                                <catalog.author-formatter
                                                                    v-bind:value="option.value" />
```

resources/views/frontend/catalog-new/index-new.blade.php
```diff
@@ -690,7 +700,10 @@ class="tw-inset-x-0 tw-box-border tw-flex tw-min-h-0 tw-flex-1 tw-flex-col tw-ov
                                             <button class="tw-m-4 tw-w-full tw-bg-sky-300 tw-p-4"
                                                 @click="dc.close">
                                                 {{ trans('item.filter.show_results') }} <span
-                                                    class="tw-font-bold">(@{{ artworks_total }})</span>
+                                                    class="tw-font-bold">(
+                                                    <catalog.number-formatter v-bind:value="artworks_total">
```
eronisko, line 704: As above...

eronisko (commented):
Looks good, just some formatting suggestions.

I'd also suggest re-naming the formatters to `catalog.formatted_number` and `catalog.author_name` as I explain [here](https://github.com/SlovakNationalGallery/webumenia.sk/pull/892#discussion_r1311191534)

resources/views/frontend/catalog-new/index-new.blade.php
FrantisekMichalSebestyen: Tryed that, for some reason self closing tags do not work 🤷

FrantisekMichalSebestyen: I will keep it as it is for now

PR #894, opened 2023-09-11
Webumenia 1923 remove pagination on admin collection/articles

FrantisekMichalSebestyen:


eronisko (approved):
Nice, thanks for running it through Prettier too.

PR #897, opened 2023-09-13
fix: change public domain status of artworks with unknown author to f…

FrantisekMichalSebestyen:


app/Item.php
```diff
@@ -578,7 +578,7 @@ public function freeFrom() {
         }
 
         if ($this->isAuthorUnknown()) {
-            return self::FREE_ALWAYS;
+            return self::UNKNOWN_AUTHOR_THRESHOLD;
```
eronisko, line 581: I would've just done 

```suggestion
            return 1940;
```

instead of defining a global constant at the top (encapsulation) but I see you're being consistent -- cool.

eronisko (approved):

PR #898, opened 2023-09-17
Artist role (autor) - parse from artist name and store in separately

rastislav-chynoransky:
https://jira.sng.sk/browse/WEBUMENIA-1533

### DIFF
--- a/app/Item.php
+++ b/app/Item.php
@@ -806,19 +806,22 @@ class Item extends Model implements IndexableModel, TranslatableContract
 
     public function syncMatchedAuthorities(\Illuminate\Support\Collection $idsWithPivotData)
     {
-        $idsToDetach = $this->authorities
-            ->filter(fn(Authority $authority) => $authority->pivot->automatically_matched)
-            ->pluck('id')
-            ->diff($idsWithPivotData->keys());
+        // Detach automatically-matched authorities that no longer match
+        $this->authorities()
+            ->wherePivot('automatically_matched', true)
+            ->whereNotIn('id', $idsWithPivotData->keys())
+            ->detach();
 
-        $this->authorities()->detach($idsToDetach);
-        $this->authorities()->syncWithoutDetaching(
-            $idsWithPivotData->map(
-                fn($pivotData) => [
-                    'automatically_matched' => true,
-                    ...$pivotData,
-                ]
-            )
-        );
+        // Update existing pivot data
+        foreach ($idsWithPivotData as $id => $pivotData) {
+            $this->authorities()->updateExistingPivot($id, $pivotData);
+        }
+
+        // Create new authorities
+        $existingIds = $this->authorities()->pluck('id');
+        $newIds = $idsWithPivotData->keys()->diff($existingIds);
+        $this->authorities()
+            ->withPivotValue('automatically_matched', true)
+            ->attach($idsWithPivotData->only($newIds));
     }
 }
--- a/app/Item.php
+++ b/app/Item.php
@@ -180,7 +180,12 @@ class Item extends Model implements IndexableModel, TranslatableContract
 
     public function authorities()
     {
-        return $this->belongsToMany(\App\Authority::class, 'authority_item', 'item_id', 'authority_id')->withPivot('role');
+        return $this->belongsToMany(
+            \App\Authority::class,
+            'authority_item',
+            'item_id',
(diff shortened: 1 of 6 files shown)


app/Item.php
```diff
@@ -798,4 +803,22 @@ public function getImageUrlAttribute()
...
+        $this->authorities()->detach($idsToDetach);
+        $this->authorities()->syncWithoutDetaching(
+            $idsWithPivotData->map(
+                fn($pivotData) => [
+                    'automatically_matched' => true,
+                    ...$pivotData,
+                ]
+            )
+        );
+    }
```
eronisko, line 823: Just to make sure I understand: this removes all automatically_matched authorities that are not present in `$idsWithPivotData` keys?

Would this work too?

```suggestion
    {
        $this->authorities()->wherePivot('automatically_matched', true)->sync(
            $idsWithPivotData->map(
                fn($pivotData) => [
                    'automatically_matched' => true,
                    ...$pivotData,
                ]
            )
        );
    }
```

If so, could it be inlined instead? (I can only see it used once)

tests/Models/ItemTest.php
```diff
@@ -184,4 +186,44 @@ protected function createFreeItem()
             'date_latest' => 1, // CE
         ]);
     }
+
+    public function testSyncMatchedAuthorities()
```
eronisko, line 190: I'm very happy to see a test here but I'm having trouble understanding what it does :(

Could you perhaps work it into a more higher-level test case for `GmuhkItemImporter`?

Maybe a test like this

```php
public function testParseRoleFromArtistName() 
{
    $data = [...];
    GmuhkItemImporter::import($data);
    $this->assertEquals('expected role', Item::first()->authorities[0]->role)
}
```

eronisko (changes requested):
I'm sure this code works but I'm having a hard time understanding it. Could you please take a look at the suggestions and/or could we have chat about this?

app/Item.php
rastislav-chynoransky: This approach won't work because the `wherePivot` condition will limit the scope of pivot records, effectively leading to inserts of duplicate pivots if there already exist some with `automatically_matched = false`

rastislav-chynoransky: And for now this change is made only for GMUHK gallery, however it will get reused in CSV importers (or any new data source).

tests/Models/ItemTest.php
rastislav-chynoransky: In fact the test makes multiple assertions and could be split into five scenarios:
- _automatically_ matched that are _present_ in newly matched are expected _to stay_ there (with updated pivot data)
- _manually_ matched that are _present_ in newly matched are expected _to stay_ there (with updated pivot data)
- _automatically_ matched that are _missing_ from newly matched are expected _to be deleted_
- _manually_ matched that are _missing_ from newly matched are expected _to stay_ there
- and finally any new matches should be created

Of course I can create a higher-level test for importing role, but the test of these relationship changes should stay on the model-level.

rastislav-chynoransky: added higher-level test https://github.com/SlovakNationalGallery/webumenia.sk/pull/898/commits/892ca5bd099db0a85634f5734ed6dc6cfecad0f0

app/Item.php
rastislav-chynoransky: @eronisko I've included your suggestion discussed via slack https://github.com/SlovakNationalGallery/webumenia.sk/pull/898/commits/84c5076013700a810f02b97e0f4d49f4fe73d474

tests/Harvest/Importers/GmuhkItemImporterTest.php
```diff
@@ -0,0 +1,47 @@
...
+            'birth_year' => null,
+            'death_year' => null,
+        ]);
+
+        $row = FakeRecordFactory::buildGmuhkItem([
+            'author' => ['Miroslav Podhrázský (fotograf)'],
+        ]);
+        $item = $this->importer->import($row, new Progress());
+        $item->refresh();
+        $this->assertEquals('fotograf', $item->authorities[0]->pivot->role);
```
eronisko, line 45: Very nice, thank you

eronisko (approved):
I've got a couple more suggestions with the global `withPivot` being a serious concern. Otherwise, this is good to go, thank you.

PR #899, opened 2023-09-20
Authority won't update after havest with new non-preferred names added

rastislav-chynoransky:
https://jira.sng.sk/browse/WEBUMENIA-2007

### DIFF
--- a/app/Harvest/Importers/AbstractImporter.php
+++ b/app/Harvest/Importers/AbstractImporter.php
@@ -8,6 +8,7 @@ use Illuminate\Database\Eloquent\Model;
 use Illuminate\Database\Eloquent\Relations\BelongsToMany;
 use Illuminate\Database\Eloquent\Relations\HasMany;
 use Illuminate\Database\Eloquent\Relations\MorphMany;
+use Illuminate\Support\Facades\DB;
 
 abstract class AbstractImporter
 {
@@ -109,8 +110,11 @@ abstract class AbstractImporter
         foreach ($relatedRows as $relatedRow) {
             $data = $this->mappers[$field]->map($relatedRow);
             $conditions = $this->getConditions($field, $data);
-            $instance = $model->$field()->firstOrNew($conditions);
-            $instance->forceFill($data);
+            $query = $model->$field();
+            foreach ($conditions as $column => $value) {
+                $query->where(DB::raw("BINARY $column"), $value);
+            }
+            $instance = $query->firstOrNew()->forceFill($data);
             $instance->save();
             $updateIds[] = $instance->getKey();
         }
@@ -139,7 +143,11 @@ abstract class AbstractImporter
         foreach ($relatedRows as $relatedRow) {
             $data = $this->mappers[$field]->map($relatedRow);
             $conditions = $this->getConditions($field, $data);
-            $relatedModel = $relatedModelClass::where($conditions)->first();
+            $query = $relatedModelClass::query();
+            foreach ($conditions as $column => $value) {
+                $query->where(DB::raw("BINARY $column"), $value);
+            }
+            $relatedModel = $query->first();
 
             if (!$relatedModel && !$allowCreate) {
                 continue;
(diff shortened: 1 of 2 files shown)


app/Harvest/Importers/AbstractImporter.php
```diff
@@ -109,7 +109,7 @@ protected function processHasMany(Model $model, $field, array $relatedRows, $all
         foreach ($relatedRows as $relatedRow) {
             $data = $this->mappers[$field]->map($relatedRow);
             $conditions = $this->getConditions($field, $data);
-            $instance = $model->$field()->firstOrNew($conditions);
+            $instance = $model->$field()->binaryFirstOrNew($conditions);
             $instance->forceFill($data);
```
eronisko, line 113: I'd like to point out two concerns with this approach (which also apply to `whereBinary` below)

1. This now says that _all_ the matching is done in the "binary" form (not sure what the right term is). Is that desirable? Wouldn't it generate false-positives?
2. If we sometimes do want the "strict" (i.e. not binary?) matching, we don't have a way to do it.

I'd suggest chaining the `where` conditions instead of build up the big `$conditions` array. I believe it would generate same results and allow us for some more flexibility. It could also be more specific to the problem at hand -- matching names.

Something like this (untested):

```suggestion
            $instance = $model->$field()->query()
            foreach ($conditions as $column => $value) {
                // Match name without diacritics
                if ($column === 'name') {
                    $instance->whereRaw("BINARY $column", $value);
                    continue;
                }
                $instance->where($column, $value);
            }
            $instance->binaryFirstOrNew()->forceFill($data);
```

what do you think?

rastislav-chynoransky: The accent/case-insenstivity works for use-cases like user search but here I think, on the contrary, it should be a strict comparison as you basically build a unique key that represents the importing record. Those data (ID's and such) usually don't change in time so if there is a 'change' it should cause creation of a new record.

eronisko: Oh, I misunderstood what `BINARY` does. Sorry & thanks.

tests/Harvest/Importers/AuthorityImporterTest.php
```diff
@@ -88,4 +88,22 @@ public function testRelatedButNotExisting()
...
+                ],
+                [
+                    'name' => ['Marak, Julius'],
+                    'prefered' => [false],
+                ],
+            ],
+        ]);
+        $authority = $this->importer->import($row, new Progress());
+        $this->assertCount(2, $authority->names);
+    }
```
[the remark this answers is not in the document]
eronisko: So this is actually here to ensure there's no regression? I.e. that these changes did not break matching by name?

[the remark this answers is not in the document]
rastislav-chynoransky: Previously, it would import just the first record and the second would be ignored (loose comparison would test positive for existence).

app/Harvest/Importers/AbstractImporter.php
eronisko: Still though, I think I'd still prefer the

```php
foreach ($conditions as $column => $value) {
    $instance->whereRaw("BINARY $column", $value);
}
```

instead of the `whereBinary` macro definition, as that seems mighty complicated. AFAIK the for loop should be just as performant and is easier to digest.

rastislav-chynoransky: @eronisko Sure, I'm okay with a simpler approach, this is a way better, thanks!

My original thought was to convert all tables to an accent-sensitive collation as we use Elasticsearch for full-text searches. We could still opt-in for different collation per use case. That's out of scope of this task though, but would be great to consider and discuss further.

Also we're currently using `utf8_slovak_ci` everywhere and since the database contains data of various languages it would certainly need at least a general collation.

eronisko: Thanks for the explanation, I wasn't aware of all this context. Looking forward to talking about it more

eronisko (approved):
Thanks again for all the changes.

PR #905, opened 2023-10-04
feat: add fuzzy seach on filter

FrantisekMichalSebestyen:


resources/js/components/catalog-new/SearchOptionsController.vue
```diff
@@ -21,10 +21,9 @@ export default {
                     )
                     .map((selected) => ({ value: selected, count: 0, checked: true })),
             ]
-
-            return optionsWithSelected.filter((option) =>
-                option.value.toLowerCase().includes(query)
-            )
+            return matchSorter(optionsWithSelected, this.search, {
+                keys: [(option) => option.value.replace(/^([^,]*),\s*(.*)$/, '$2 $1')],
```
FrantisekMichalSebestyen, line 25: We might want to extract this somehow, so it's not in 2 places at once
In AuthorFormatter and here

FrantisekMichalSebestyen: WDYT? @eronisko

eronisko: Would `keys: ['value']` generate worse results?

PR #920, opened 2023-10-31
[harvest] gmuhk default authority role

rastislav-chynoransky:
https://jira.sng.sk/browse/WEBUMENIA-1533

### DIFF
--- a/app/Harvest/Importers/GmuhkItemImporter.php
+++ b/app/Harvest/Importers/GmuhkItemImporter.php
@@ -47,7 +47,7 @@ class GmuhkItemImporter extends AbstractImporter
             ->map->first()
             ->mapWithKeys(
                 fn(Authority $authority, $author) => [
-                    $authority->id => ['role' => AuthorityMatcher::parse($author)['alt_name']],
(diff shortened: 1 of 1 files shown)


app/Harvest/Importers/GmuhkItemImporter.php
```diff
@@ -47,7 +47,7 @@ public function import(array $row, Progress $result) {
             ->map->first()
             ->mapWithKeys(
                 fn(Authority $authority, $author) => [
-                    $authority->id => ['role' => AuthorityMatcher::parse($author)['alt_name']],
+                    $authority->id => ['role' => AuthorityMatcher::parse($author)['alt_name'] ?? 'autor/author'],
```
eronisko, line 50: The role is being parsed out of `alt_name`? Is this right?

rastislav-chynoransky: For now yes, this GMUHK data has the `role` saved where other data sources have `alt_name`.
We're going to refactor the matcher/parser soon but that will require few more steps – matcher will need to understand the format of `Item::$author`.

eronisko (approved):

PR #929, opened 2023-12-07
[import] mg foto importer

rastislav-chynoransky:
https://jira.sng.sk/browse/MG-13


rastislav-chynoransky:
tests are failing due to missing translations. we'll add them before merging

app/Importers/AbstractMgImporter.php
```diff
@@ -0,0 +1,75 @@
+<?php
+
+namespace App\Importers;
+
+class AbstractMgImporter extends AbstractImporter
```
eronisko, line 5: Not a blocker but could we make this into a trait rather than a second layer of abstract classes? (composition over inheritance)

tests/Importers/MgFotoImporterTest.php
```diff
@@ -0,0 +1,331 @@
...
+                'papír',
+                'M',
+                [
+                    'cs' => 'papír/matný',
+                    'sk' => 'papier/matný',
+                    'en' => 'paper/matte',
+                ],
+            ],
+        ];
+    }
```
eronisko, line 73: Sorry, I still don't get it -- why do we need the dataproviders here?

rastislav-chynoransky: It's the same test with different data. Would you rather see it as multiple tests?

eronisko: It's fine in a single test, I just don't understand why we need the data providers. Or maybe I don't understand what the data providers do? Not an objection, but a question.

eronisko, line 73: I've _finally_ gotten what the dataproviders do :D 

For what it's worth, I find this more legible and straightforward. Not much value in DRYing this up IMHO.

```suggestion
    public function testMedium()
    {
        $this->importData([
            'Material' =>'papír',
            'Povrch' => null,
        ]);

        $item = Item::find('CZE:MG.rada_s_123-lomeni_s');
        $this->assertEquals('papír', $item->translate('cs')->medium);
        $this->assertEquals('papier', $item->translate('sk')->medium);
        $this->assertEquals('paper', $item->translate('en')->medium);
        
        $this->importData([
            'Material' =>'papír',
            'Povrch' => 'M',
        ]);

        $item = Item::find('CZE:MG.rada_s_123-lomeni_s');
        $this->assertEquals('papír/matný', $item->translate('cs')->medium);
        $this->assertEquals('papier/matný', $item->translate('sk')->medium);
        $this->assertEquals('paper/matte', $item->translate('en')->medium);
    }
```

eronisko (approved):
LGTM, thank you

tests/Importers/MgFotoImporterTest.php
rastislav-chynoransky: The dataProvider here helps to keep the tests isolated though.

PR #959, opened 2024-02-29
Enable search and aggregations combining authors with and without authority_id 

eronisko:
# Description

### New `authors` field for `/api/v1/items`
Lists both authors with and without Authority records.
```jsonc
//...
"authors": [
  {
    "name": "Vodrážka, Jaroslav",
    "authority": {
      "id": "11436"
    },
    "name_formatted": "Jaroslav Vodrážka"
  },
  {
    "name": "Viedenský maliar z konca 18. storočia",
    "authority": null,
    "name_formatted": "Viedenský maliar z konca 18. storočia"
  }
]
```
Note that this is currently different from the `authors` field in `/api/v2/items/<id>`
```jsonc
//...
"authors": [
  "Vodrážka, Jaroslav",
  "Viedenský maliar z konca 18. storočia"
]

```

### `/api/v1/items/aggregations` also include a new `authors` term
listing both the author and their Authority ID:

```jsonc
// GET /api/v1/items/aggregations?terms[authors]=authors
{
  "authors": [
    {
      "name": "Vodrážka, Jaroslav",
      "authority": {
        "id": "11436"
      },
      "count": 6
    },
    {
      "name": "Nemecký maliar zo začiatku 19. storočia",
      "authority": null,
      "count": 2
    }
  ]
}
```

This will also list multiple entries for authors with the same name but different authority IDs

###  Search by `author.name` and `author.authority.id` is now supported
When using both, the search is executed as an 'or' query.

### Elasticsearch is upgraded to 7.17.3
In order to enable aggregations by multiple terms

## Type of change

# Checklist:

### DIFF
--- a/app/Http/Controllers/Api/V1/ItemController.php
+++ b/app/Http/Controllers/Api/V1/ItemController.php
@@ -114,6 +114,29 @@ class ItemController extends Controller
         $aggregationsQuery = [];
 
         foreach ($terms as $aggregationName => $field) {
+            if ($field === 'authors') {
+                $aggregationsQuery[$aggregationName]['aggregations']['filtered'] = [
+                    'nested' => [
+                        'path' => 'authors',
+                    ],
+                    'aggs' => [
+                        'authors' => [
+                            'multi_terms' => [
+                                'terms' => [
+                                    [
+                                        'field' => 'authors.name.keyword',
+                                    ],
+                                    [
+                                        'field' => 'authors.authority.id.keyword',
+                                        'missing' => '',
+                                    ],
+                                ],
+                            ],
+                        ],
+                    ],
+                ];
+                continue;
+            }
             $aggregationsQuery[$aggregationName]['aggregations']['filtered']['terms'] = [
                 'field' => $field,
                 'size' => $size,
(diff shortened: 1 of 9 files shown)


app/Http/Controllers/Api/V1/ItemController.php
```diff
@@ -265,6 +313,17 @@ protected function createQueryBuilder($q, $filter)
                 $builder->filter($colorQuery);
                 continue;
             }
+
+            if ($field === 'authors.name' || $field === 'authors.authority.id') {
+                $authorsBuilder ??= Query::bool();
+                $authorsBuilder->should(
```
rastislav-chynoransky, line 319:
```suggestion
                $authorsBuilder->must(
```

Just a question. Why is this doing a union instead of an intersection? Is there a specific use case for that?

rastislav-chynoransky (approved):
👏 Very nice! Thanks for this.

app/Http/Controllers/Api/V1/ItemController.php
eronisko: For posterity: It's doing a union because I want to be able to see both authors with and without authority ID in the search list. When I pick a 'authority' author, an ID is added to the filter, when I pick a 'non-authority' author, their name is added.

eronisko:
I've decided to merge this next week when I can also run a migration of the Elasticsearch index

PR #986, opened 2024-04-01
Basic collections API

rastislav-chynoransky:
https://jira.sng.sk/browse/MG-53

Run dev server with `PHP_CLI_SERVER_WORKERS=2 php artisan serve` to be able to serve collections index concurrently.

**TODO**
1. parse MG urls
2. generate image thumbnails
3. add global frontend scope for collections
4. restrict permissions for MG admin users


app/Repositories/CollectionRepository.php
```diff
@@ -0,0 +1,45 @@
...
+        $responses = Http::pool(
+            fn(Pool $pool) => $collections
+                ->filter(fn(Collection $collection) => $collection->filter_items_url)
+                ->map(
+                    fn(Collection $collection) => $pool
+                        ->as($collection->id)
+                        ->withHeaders(['X-Frontend' => Frontend::get()->value])
+                        ->get($collection->filter_items_url)
+                )
+        );
```
eronisko, line 29: I don't know if this is a good way to do it in terms of design.

I'd suggest either implementing the conversion logic from URL to a filter/query (rather than the API URL) and then get them from ES directly. The other alternative would be just not show item counts/items in the `index` response. Otherwise we're running into the risk of DoS-ing our own API.

phpunit.xml
```diff
@@ -12,10 +12,9 @@
   </coverage>
   <php>
     <env name="APP_ENV" value="testing"/>
+    <env name="APP_URL" value="http://localhost"/>
     <env name="BCRYPT_ROUNDS" value="4"/>
     <env name="CACHE_DRIVER" value="array"/>
-    <!-- <env name="DB_CONNECTION" value="sqlite"/> -->
-    <!-- <env name="DB_DATABASE" value=":memory:"/> -->
```
eronisko, line 18: FYI, this is standard boilerplate: https://github.com/laravel/laravel/blob/10.x/phpunit.xml

tests/Feature/Api/Collections/IndexTest.php
```diff
@@ -0,0 +1,81 @@
...
+        $response->assertJsonCount(1, 'data')->assertJsonPath('data.0', [
+            'id' => $collection->id,
+            'name' => $collection->name,
+            'text' => $collection->text,
+            'filter_items_url' => route('api.v1.items.index', [
+                'filter' => [
+                    'author' => ['author-1'],
+                ],
+            ]),
+            'filter_items_count' => 1,
```
eronisko, line 58: IMHO I think the ideal API would just give me
```suggestion
            'id' => $collection->id,
            'name' => $collection->name,
            'text' => $collection->text,
            'items' => [$item1, $item2] // items list/resouce,
```
And I shouldn't need to care if the items are coming from URL or a list of IDs.

eronisko (commented):
I have some bold suggestions here :D let me know what you think

tests/Feature/Api/Collections/IndexTest.php
rastislav-chynoransky: It would be better to retrieve paginated items from a separate endpoint, such as `api/collection/{id}/items`. However, at this point, we only need to implement filtered collections for MG. Therefore, the existing endpoint `api/v1/items` could suffice.

app/Repositories/CollectionRepository.php
rastislav-chynoransky: As discussed on the call we gonna drop this functionality.

app/Http/Controllers/Api/CollectionItemController.php
```diff
@@ -0,0 +1,31 @@
...
+        if (request()->has('size')) {
+            $parameters['size'] = request()->integer('size');
+        }
+
+        if (request()->has('page')) {
+            $parameters['page'] = request()->integer('page');
+        }
+
+        $request = Request::create(route('api.v1.items.index', $parameters));
+        return app()->handle($request);
```
eronisko, line 29: Nice trick 👍

tests/TestCase.php
eronisko, line 1: 🎖️

tests/Unit/CollectionTest.php
```diff
@@ -0,0 +1,34 @@
...
+    public function testItemFilterFromMoravskaGalerieUrl()
+    {
+        $collection = Collection::factory()->make([
+            'url' => 'https://sbirky.moravska-galerie.cz/?work_type=maliarstvo|fotografia',
+        ]);
+
+        $expected = [
+            'work_type' => ['maliarstvo', 'fotografia'],
+        ];
+        $this->assertEquals($expected, $collection->item_filter);
```
eronisko, line 32: Just, wondering, what's the `item_filter` for? Some presentational purpose?

phpunit.xml
```diff
@@ -12,6 +12,7 @@
   </coverage>
   <php>
     <env name="APP_ENV" value="testing"/>
+    <env name="APP_URL" value="http://localhost"/>
```
eronisko, line 15:
```suggestion
```

Probably no longer needed?

eronisko (approved):
Look great! Thank you.

tests/Unit/CollectionTest.php
rastislav-chynoransky: Yes, we’ll display it in the collection detail.

PR #988, opened 2024-04-08
[collection] add frontend scope

rastislav-chynoransky:
https://jira.sng.sk/browse/MG-53

### DIFF
--- a/lang/sk/permissions.php
+++ b/lang/sk/permissions.php
@@ -1,0 +1,8 @@
+<?php
+
+return [
+    'can_administer' => 'Administrovanie',
+    'can_edit' => 'Editovanie',
+    'can_publish' => 'Publikovanie',
+    'can_import' => 'Import',
+];
--- a/lang/sk/permissions.php
+++ b/lang/sk/permissions.php
@@ -1,7 +1,7 @@
 <?php
 
 return [
-    'can_administer' => 'Administrovanie',
+    'can_administer' => 'Administrácia',
     'can_edit' => 'Editovanie',
     'can_publish' => 'Publikovanie',
     'can_import' => 'Import',
--- a/app/Collection.php
+++ b/app/Collection.php
@@ -46,6 +46,7 @@ class Collection extends Model implements TranslatableContract
 
     protected $casts = [
         'published_at' => 'datetime',
+        'frontends' => 'array',
     ];
 
     public function items()
--- a/app/Http/Controllers/CollectionController.php
+++ b/app/Http/Controllers/CollectionController.php
@@ -62,6 +62,8 @@ class CollectionController extends Controller
         if ($v->passes()) {
             $collection = new Collection();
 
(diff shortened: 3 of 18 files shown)


eronisko:
I'm just writing these down so that I don't forget when we discuss them.

- Right now, I believe the `editor` role has the permission to publish. I'd keep using this role instead of creating a new `publisher` role
- I would also keep `role` as a singular attribute, so that we don't mix up role-based and permission based access models. The frontend would narrow down permissions for non-administrator roles (e.g. editors)
- I think editors could be restricted to a single frontend for now (just to keep things simpler when filtering -- it's a minor sugestion)
- When creating (or editing) a new editor user, I am required to specify a frontend 
- When viewing as an editor, I can only see resources from my frontends
- When viewing as an administrator, I can see all resources
- When editing as an administrator, I can specify frontends for a resource (have to pick at least one)
- When editing as an editor, I'm not able to change frontends of the resource
- Only administrators would have the options to change frontends when creating/editing resources (collections, articles, ...). - When creating as an editor, the resource will automatically my frontend
- When creating as an administrator, I will be able to specify the frontends (but `['webumenia']` is probably a good default)

rastislav-chynoransky:
> * Right now, I believe the `editor` role has the permission to publish. I'd keep using this role instead of creating a new `publisher` role

As discussed with @mcudrnak it seems that we don't want editors to have permissions to publish.

> * I would also keep `role` as a singular attribute, so that we don't mix up role-based and permission based access models. The frontend would narrow down permissions for non-administrator roles (e.g. editors)

I would prefer to switch to multiple roles, because a user should be to edit/publish/import and there is no inherent hierarchy in it.

> * I think editors could be restricted to a single frontend for now (just to keep things simpler when filtering -- it's a minor sugestion)

Filtering in the admin panel is currently based on ownership of the resource. 

> * When creating (or editing) a new editor user, I am required to specify a frontend

👍 

> * When viewing as an editor, I can only see resources from my frontends

We can also keep the filter based on ownership, I think there'e no need to change this now.

> * When viewing as an administrator, I can see all resources

👍 

> * When editing as an administrator, I can specify frontends for a resource (have to pick at least one)

👍 

> * When editing as an editor, I'm not able to change frontends of the resource

👍 

> * Only administrators would have the options to change frontends when creating/editing resources (collections, articles, ...). - When creating as an editor, the resource will automatically my frontend

This is a great point, I'm all for it!

> * When creating as an administrator, I will be able to specify the frontends (but `['webumenia']` is probably a good default)

👍

lang/sk/permissions.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+return [
+    'can_administer' => 'Administrovanie',
```
eronisko, line 4:
```suggestion
    'can_administer' => 'Administrátor',
```

eronisko (approved):
Looks good! Thank you

PR #792, opened 2022-11-21
disable throttling for authenticated users

rastislav-chynoransky:
https://jira.sng.sk/browse/NSNG-84

Source of `SwitchGuard` middleware:
https://dannyherran.com/2022/01/laravel-sanctum-optional-route-authentication-guard/

PR #997, opened 2024-04-17
Rate limiter whitelist of IP address

rastislav-chynoransky:
As suggested in #792

### DIFF
--- a/config/app.php
+++ b/config/app.php
@@ -227,8 +227,6 @@ return [
     'iip_public' => env('IIP_PUBLIC', 'https://img.webumenia.sk'),
 
     'rate_limiter' => [
-        'whitelisted_ip_addresses' => env('RATE_LIMITER_WHITELISTED_IP_ADDRESSES')
-            ? explode(',', env('RATE_LIMITER_WHITELISTED_IP_ADDRESSES'))
-            : [],
+        'whitelisted_ip_addresses' => ['195.210.28.242'],
     ],
 ];
--- a/config/app.php
+++ b/config/app.php
@@ -225,4 +225,10 @@ return [
 
     'iip_private' => env('IIP_PRIVATE', 'http://127.0.0.1:8002'),
     'iip_public' => env('IIP_PUBLIC', 'https://img.webumenia.sk'),
+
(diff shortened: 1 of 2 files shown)


config/app.php
```diff
@@ -225,4 +225,10 @@
 
     'iip_private' => env('IIP_PRIVATE', 'http://127.0.0.1:8002'),
     'iip_public' => env('IIP_PUBLIC', 'https://img.webumenia.sk'),
+
+    'rate_limiter' => [
+        'whitelisted_ip_addresses' => env('RATE_LIMITER_WHITELISTED_IP_ADDRESSES')
+            ? explode(',', env('RATE_LIMITER_WHITELISTED_IP_ADDRESSES'))
+            : [],
```
eronisko, line 232: Is there an actual need for the configurability? Could we just list the IPs here?

rastislav-chynoransky: It would be handy to be able to add our own IP addresses.

eronisko: Do you mean for development? Maybe we could disable it based on app environment then?

eronisko (approved):
I don't have any issues with the implementation, but I'm a little concerned that the whitelisting is potentially needed for development. In my opinion, the rate limiter should be a safety net, not a mode of operation.

My guess is that either
- the rate limiter is too strict, or
- the client implementation is not optimal, or
- the API needs a design adjustment

In any case, from an architecture perspective, I don't think the assumption that "nobody should call this so often, except for us" is 100% sound. The "whitelisted" calls are ultimately being triggered by visitors, so still outside of our control. In fact, for those apps -- that might potentially generate lots of requests -- we're losing that safety mechanism altogether.

My suggestion would be to increase the rate limit for everyone to a level we're comfortable with.

cc @igor-kamil @FrantisekMichalSebestyen

rastislav-chynoransky:
When using SSR frontends the requests will come from our own server, that's why we need to make a whitelist here. And if high rates are an issue to implement rate limiter on the caller side.

igor-kamil (commented):
for now just turn it off. in case of troubles, we can back to it

PR #1012, opened 2024-05-16
[import] trim all whitespaces in mg importers

rastislav-chynoransky:
https://jira.sng.sk/browse/MG-51


igor-kamil (approved):
LGTM. however I guess it makes sense to trim all imported data (we don't want trailing spaces). 
what about to use [stream filters](https://csv.thephpleague.com/9.0/connections/filters/)? (just an idea...)

rastislav-chynoransky:
You're right, I think we can apply it to all fields. We already got a method of doing exactly that. The stream filter would have to be registered only for this use case.

PR #1017, opened 2024-06-19
[item] allow viewing of all items in api v2

rastislav-chynoransky:
https://jira.sng.sk/browse/MG-100

### DIFF
--- a/app/Http/Controllers/Api/V2/ItemController.php
+++ b/app/Http/Controllers/Api/V2/ItemController.php
@@ -17,14 +17,18 @@ class ItemController extends Controller
 
     public function index(Request $request)
     {
-        $size = $request->input('size', 15);
+        $validated = $request->validate([
+            'ids' => ['array'],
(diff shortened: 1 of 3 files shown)


app/Http/Controllers/Api/V2/ItemController.php
```diff
@@ -17,16 +17,13 @@ public function show($id)
 
     public function index(Request $request)
     {
-        $ids = $request->input('ids');
-
-        if (!$ids || !is_array($ids)) {
-            return response()->json(['error' => 'Invalid input'], 400);
-        }
```
igor-kamil, line 24: just saying - I would like to keep validation whether `ids` is array. it doesn't have to be required.

igor-kamil (approved):
ok. just a tiny note. not necessary to implement. can be merged as it is.

igor-kamil (approved):
very nice! 👌

PR #1021, opened 2024-08-16
Fix sortable in collections

rastislav-chynoransky:
https://jira.sng.sk/browse/WEBUMENIA-2090


igor-kamil (approved):
👌 thanks for the fix

PR #1024, opened 2024-08-20
Use gate instead of user attribute

rastislav-chynoransky:
https://jira.sng.sk/browse/MG-108


rastislav-chynoransky:
@igor-kamil bump

igor-kamil (approved):

PR #1039, opened 2024-11-28
Webumenia 2109 search by ID and identifier

igor-kamil:
# Description

This reflect the common use-case when colleagues search by identifier (or directly via ID).

## Type of change

### DIFF
--- a/app/Http/Controllers/Api/V1/ItemController.php
+++ b/app/Http/Controllers/Api/V1/ItemController.php
@@ -41,8 +41,6 @@ class ItemController extends Controller
         'box',
         'location',
         'location.tree',
-        'identifier',
-        'id',
     ];
 
     private $rangeables = ['date_earliest', 'date_latest', 'additionals.order'];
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/Api/V1/ItemController.php
```diff
@@ -41,6 +41,8 @@ class ItemController extends Controller
         'box',
         'location',
         'location.tree',
+        'identifier',
+        'id',
```
rastislav-chynoransky, line 45: Is this necessary?

igor-kamil: actually not... just in case you want to filter by them directly. I thought it might be useful. but definitely not necessary.

rastislav-chynoransky: Those should be unique so it's basically covered by detail route. I wouldn't add it until we have a use case for it.

rastislav-chynoransky (approved):

PR #1042, opened 2024-12-14
[item] Add command for dumping untranslated attributes

rastislav-chynoransky:


igor-kamil (approved):
looks and works nicely

